PR #5, opened 2020-11-23
Feature/add demo example

helloimela:
**Describe pull-request**  
Add Angular demo example and simple HTML

**How to test**  
1. open angular demo folder, npm i,npm start
2. To test HTML example, run http-start on sdds folder, then go to http://127.0.0.1:8080/demo/HTML/


mmexvr:
Maybe we should add for the HTML that you need to install HTTP-SERVER package (globally or locally)

demo/angular/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "zone.js": "~0.9.1"
+  },
+  "devDependencies": {
+    "@angular-devkit/build-angular": "~0.800.6",
+    "@angular/cli": "~8.0.6",
+    "@angular/compiler-cli": "^8.0.3",
+    "@angular/language-service": "^8.0.3",
+    "@scania/components": "file:../scania-components-4.2.0.tgz",
+    "@scania/theme-light": "file:../scania-theme-light-1.2.0.tgz",
+    "@sdds/components": "file:../sdds-components-4.2.0.tgz",
```
mmexvr, line 39: Had the remove this line, since that packages doesn't exist @sdds/...

This is replaced anyway with the new package

mmexvr: Another way you can do, you can also run it with npm link only and not have it in the package.json file at all you need to build and link both then.

mmexvr (approved):

Issue #31, opened 2020-12-22
Improvement - Add storybook story for grid


mmexvr:
**Describe current situation**  
_Describe what the current situation and what problem it is causing for this improvement_

There is no storybook story for the grid system

**Describe the improvement**  
_A clear and concise description of what kind of improvement you have in mind._

Add a storybook story for the whole gridsystem
- Regular
- Push
- Gutterless
- No padding
- Offset
- column options
- ...

- Spacing story added

**If applicable, please provide steps to reproduce**
_Steps to reproduce the behavior:_  
1. Run storybook
2. Look in menu, no story for grid

AB#173

Issue #18, opened 2020-12-09
Add new spacing value


helloimela:
$spacing-layout-72 needs to be added to typography package

AB#191

mmexvr:
Wasn't added as a css variable #38 fixes it

mmexvr:
Fixed in #38

PR #38, opened 2021-01-07
Added story for grid, spacing, new spacing option and more...

mmexvr:
**Describe pull-request**  
- Added grid story
- Added new spacing option 72px css var
- Added spacing story
- Fixed some column issues in color story for mobile
- Added controls for the stories
- Expanded control information for stories
- Renamed typography story 

**Solving issue**  
fixes: #31 
fixes: #18 (bug)

**How to test**  
1. Run storybook
2. Check grid story
3. Try out different options, Enable addons in the settings of storybook, top left
4. Check spacing

**Additional context**  
- Examples for grid might be a bit too complicated (?)
- Need to use the controls more :+1:

### DIFF
--- a/components/.storybook/main.js
+++ b/components/.storybook/main.js
@@ -7,6 +7,7 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-notes/register",
-    "@storybook/addon-controls"
+    "@storybook/addon-controls",
+    '@storybook/addon-actions'
   ]
 }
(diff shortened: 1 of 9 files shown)


helloimela:
The example is a bit confusing maybe because the color that blends with the background (white for column background), so it's hard to identify which one is column, which one is spacing. It is unclear what is red is, what is white is. So, maybe add color background for both spacing & column. If you see the example below, we can see clearly the box/column because they have different colour compared to background.

image

image

helloimela:
Can we add a defult text to the coloumn. I think it makes it clearer. Right now in basic example, there's no default text

image


With text, it looks understandable. In addition, we don't need to add all 16 / 8 columns in the example. Instead, we can add "real life" example, by only adding 3 column or 4 column, and add text "2 of 4" or something like that. An idea would be, we can add a text outside the columns something like: .xxlg, .xlg, .lg max column = 16, .md max column = 8, ...." and so on.

image

mmexvr:
Create simple examples, remove the columns options (range)

helloimela (approved):

PR #69, opened 2021-02-04
Feature/link

helloimela:
- Add link style on <a> element and class .sdds-link
- Disable underline on button element that is used with <a class='sdds-btn...>

How to test:
- See storybook and change option for disable to see disable state
- Figma >> UI Kit (Global) >> Link or Figma >> Components >> Link


components/src/components/button/button.stories.js
```diff
@@ -62,10 +62,10 @@ const ButtonTemplate = ({size, btnType, fullbleed, text='Button', withIcon, disa
 
   return `
   <c-theme name="scania" global="true"></c-theme>
-  <button class="sdds-btn sdds-btn-${btnType} ${sizeValue} ${fbClass} ${disabled} ${onlyIconCss}" ${inlineStyle}>
+  <a href="#" class="sdds-btn sdds-btn-${btnType} ${sizeValue} ${fbClass} ${disabled} ${onlyIconCss}" ${inlineStyle}>
```
mmexvr, line 65: Maybe keep button element for the button story, since we have a story for link?

components/src/components/link/link.scss
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  &:visited{
+    color: var(--sdds-link-visited);
+  }
+
+  &:focus{
+    color: var(--sdds-link-focus);
+    text-decoration: none;
+    border: var(--sdds-link-focus-border);
```
mmexvr, line 26: Instead of border, change to outline. Otherwise it will jump around in focused state

mmexvr (changes requested):

mmexvr (approved):

niklasros (approved):

PR #72, opened 2021-02-05
Feature/button test

iths-jonathan-engman:
**Describe pull-request**  
--Completed the rearrangement of the button components tasks, into the standard button component folder. 
--Implemented the code needed for button.stories.js, combined with the tsx-file.
--Made sdds-button in to a functional component with props.
--sdds-button now with working props, as in text, type, size, disabled and fullbleed.
--Added styling on slot, now targeting the slotname.
--Fullbleed is now working setting the class correctly
--Type is now working setting the class correctly
--Disable now a toggle option in the control
--Added a custom textfield for the button
--Added a toggle option for the icon
--Removed the toggle option for the only icon

**Solving issue**  

**How to test**  
Go to storybook
->Components/Button
->SDDS Button

**Screenshots**  

**Additional context**  
This code is for a educational purpose


components/src/components/button/button-component.tsx
```diff
@@ -0,0 +1,19 @@
+import { Component, h, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'test-btn',
```
mmexvr, line 4: I would recommend to maybe rename it

c-button

c = for component 
button = what the component is

components/src/components/button/button-component.tsx
```diff
@@ -0,0 +1,26 @@
+import { Component, h, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'sdds-button',
+  styleUrl: 'button.scss',
+  shadow: true,
+})
+
+export class CButton {
```
helloimela, line 9: Change class to Button, since we use different prefix now

components/src/components/button/button-component.tsx
```diff
@@ -0,0 +1,26 @@
...
+  @Prop() type: string;
+  @Prop() size = "";
+  @Prop() disabled: string;
+  @Prop() fullbleed: boolean;
+
+  render() {
+    return (
+      <button class={`sdds-btn sdds-btn-${this.type} sdds-btn-${this.size} ${this.disabled} ${this.fullbleed ? 'sdds-btn-fullbleed' : ''}`}>
+          {this.text}
+          <span class='sdds-btn-icon'>
```
helloimela, line 20: Investigate solution if button does not have icon, span sdds-btn-icon should not taking space to the right

components/src/components/button/button.stories.js
```diff
@@ -65,7 +65,14 @@ const ButtonTemplate = ({size, btnType, fullbleed, text='Button', withIcon, disa
...
     ${iconHTML}
-  </button>
+  </button>   
+  `
+};
+
+const ComponentBtn = () => {
+  return `
+  <c-theme name="scania" global="true"></c-theme>
+  <sdds-button type="primary" size="md" disabled="" fullbleed="true" text="Button"><c-icon name='scania-cross'></c-icon></sdds-button>
```
helloimela, line 75: Provide storybook args, so that when we change value in control, the button will change according to the given value. Something like size=${size}. See example on ButtonTemplate

helloimela (changes requested):

components/src/components/button/button.scss
```diff
@@ -104,4 +104,10 @@ $props : (
     }
   }
   
+}
+
+::slotted( [slot="icon"] ) {
+  margin-left: var(--sdds-spacing-element-32);
+  width: 16px;
```
mmexvr, line 111: Maybe you can use spacing variables here also?  --sdds-spacing-element-16 for example

helloimela (approved):
great job!

PR #85, opened 2021-03-01
AB#474 Feature/dropdown

helloimela:
**Describe pull-request**  
- Added dropdown styling
- Added dropdown interaction

**Solving issue**  
Fixes: [AB#474](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/474)

**How to test**  
- Check Figma for styling guidelines
- Check if dropdown works in different browser
- Basic dropdown (not include different types)

- [ ] Label inside
- [ ] Label outside
- [ ] Error state
- [ ] Helper text

NOT INCLUDED in this PR
-  Multi select
-  Filter
-  Nested

### DIFF
--- a/components/src/components/button/readme.md
+++ b/components/src/components/button/readme.md
@@ -1,0 +1,21 @@
+# sdds-button
+
+
+
+<!-- Auto Generated Below -->
+
+
+## Properties
+
+| Property    | Attribute   | Description | Type      | Default     |
+| ----------- | ----------- | ----------- | --------- | ----------- |
+| `disabled`  | `disabled`  |             | `boolean` | `undefined` |
+| `fullbleed` | `fullbleed` |             | `boolean` | `undefined` |
+| `size`      | `size`      |             | `string`  | `""`        |
+| `text`      | `text`      |             | `string`  | `undefined` |
+| `type`      | `type`      |             | `string`  | `undefined` |
+
+
+----------------------------------------------
+
+*Built with [StencilJS](https://stenciljs.com/)*
--- a/components/src/components/dropdown/dropdown-option.tsx
+++ b/components/src/components/dropdown/dropdown-option.tsx
@@ -1,0 +1,59 @@
+import {
+    Component, h, Prop, State, Element, Host,
+    Event, EventEmitter, Listen
+  } from '@stencil/core';
+  
+  @Component({
+    tag: 'sdds-dropdown-option',
+    styleUrl: 'dropdown.scss',
+    shadow: true,
+  })
+  export class DropdownOption {
+    
+    @Element() host: HTMLElement;
+
+    @State() selected:boolean=false;
+    @State() optionLabel:string;
+
+    /** Value is a unique string that will be used for application logic */
+    @Prop() value:string;    
+
+    @Event({
+      eventName: 'selectOption',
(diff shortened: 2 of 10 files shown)


niklasros:
I can not see the dropdown options when it's clicked on Firefox and Safari. It is visible in Chrome.

dropdown-screenshot

niklasros:
The label outside doesn't position the label outside when it's default. Only when an option has been selected, the label is visible outside the dropdown.

_Default_
label-outside-default

_Selected_
label-outside-selected

niklasros:
The label text looks very small on Storybook, compared to the Figma design. Also when it's compared with text fields in Storybook. When I check the size of it, it says 14px, but I think something makes the text smaller than what it is actually is. 

label-text

helloimela:
@niklasros it's fixed now 😃

niklasros:
**Q:** Should we write out the text "Label text" and "Dropdown text" in Storybook so that it easier for the developers to see which text is which on the different types? 

For instance, on "Label inside" the label text is labeled _Select option_ and on "Label outside" the label text is labeled _Label text_ and the dropdown text _Select option_.

helloimela:
> **Q:** Should we write out the text "Label text" and "Dropdown text" in Storybook so that it easier for the developers to see which text is which on the different types?
> 
> For instance, on "Label inside" the label text is labeled _Select option_ and on "Label outside" the label text is labeled _Label text_ and the dropdown text _Select option_.

@niklasros  now I have changed it to be more consistent.  In the code, _Select dropdown_ will be called `placeholder` and _Label text_ is `label`

niklasros:
> **Q:** Should we write out the text "Label text" and "Dropdown text" in Storybook so that it easier for the developers to see which text is which on the different types?
> 
> For instance, on "Label inside" the label text is labeled _Select option_ and on "Label outside" the label text is labeled _Label text_ and the dropdown text _Select option_.

It looks great now. My only comment is that on "Label inside" the placeholder text is set as Select option and label text as Label text as default, which makes it looks like two different texts instead of one (as the text field) when the user has done a selection.

I understand that the text is editable in Storybook for instance, so maybe it's not an issue. It just looks a tad off when interacting with the component in Storbyook when its default and you haven't edited the texts :)

niklasros (approved):

mmexvr:
Maybe we should add the possibility to deselect a value? 

Pick option 1 then deselect it to return to default state

helloimela:
> Maybe we should add the possibility to deselect a value?
> 
> Pick option 1 then deselect it to return to default state

@mmexvr never seen that pattern in other dropdown, maybe we can ask UX for clarification?

mmexvr:
When I have selected an option and reopen, the option is still selected. 
But when I click outside the dropdown then click again the option inside the menu isn't selected with the checkmark. 

Maybe improvement for next iteration?

helloimela:
@mmexvr good catch! Added an issue here https://github.com/scania-digital-design-system/sdds/issues/90
Can works on this next sprint, we can merge this one first if there's no other serious bug?

mmexvr:
Great work!

mmexvr (approved):

PR #86, opened 2021-03-02
AB#298 Feature/tooltip

iths-jonathan-engman:
**Describe pull-request**  
--Tooltip component with control for changing the borders, also with the default state.
--Including adding custom text.
--Size adapts to filltext

**Currently not working**
--Tooltip is still in progress

**How to test**  
--Tooltip -> basic.

Feel free to try out different positions by changing the placement value in tooltip.stories.js via sdds-tooltip.
placement(s):
  | 'auto'
  | 'auto-start'
  | 'auto-end'
  | 'top'
  | 'top-start'
  | 'top-end'
  | 'bottom'
  | 'bottom-start'
  | 'bottom-end'
  | 'right'
  | 'right-start'
  | 'right-end'
  | 'left'
  | 'left-start'
  | 'left-end'

**Additional context**


niklasros:
Is it possible to get an example in Storybook on how the transition fade-in/out looks like when hovering an element that triggers a tooltip? Would be great to have that defined, like a default quick fade-in/out transition for all tooltips.

iths-jonathan-engman:
> Is it possible to get an example in Storybook on how the transition fade-in/out looks like when hovering an element that triggers a tooltip? Would be great to have that defined, like a default quick fade-in/out transition for all tooltips.

Commenting this for transparency.
I talked to Niklas regarding the comment, and he wishes to see a live example of a working tooltip attached to a element.

components/src/components/tooltips/tooltip.tsx
```diff
@@ -0,0 +1,20 @@
+import { Component, h, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'sdds-tooltip',
+  styleUrl: 'tooltip.scss',
+  shadow: true,
+})
+
+export class Tooltips {
```
helloimela, line 9: A single component, Class should be named Tooltip. And to be consistent with tag name as well

niklasros:
I think we need to change the structure a bit for the tooltip.

**Option Basic** 
The possibility to render all the tooltip types (as your previous solution)

**Option Transition**
Example of how the tooltip is animated in. On this option, it's enough to only show one tooltip type. As it is now, it's a bit confusing since the hover can show all the types, and then positions the tooltip in the wrong place.

tooltip-hover-example

niklasros:
Add a quick fade-in transition for the tooltip, so that the transition becomes a bit softer.

components/src/components/tooltips/tooltip.stories.js
```diff
@@ -0,0 +1,50 @@
...
+    background: #16417F;
+    color: white;
+    padding: 2rem;
+    border-radius: 1rem;
+  }
+  </style>
+
+  <c-theme name="scania" global="true"></c-theme>
+
+  <div class="target-wrap">
```
helloimela, line 38: Try to avoid your own styling on target element and use sdds component such as sdds-button to be used as an example in storybook

components/src/components/tooltips/tooltip.scss
```diff
@@ -0,0 +1,99 @@
...
+.sdds-tooltip {
+  @include type-style('detail-05');
+  color: var(--sdds-tooltip-colour);
+  background-color: var(--sdds-tooltip-background);
+  border-radius: 1rem;
+  padding: 2rem;
+  position: absolute;
+  word-wrap: break-word;
+  max-width: 48rem;
+  top: 20rem;
```
helloimela, line 71: If looking at the UI guidelines in Figma & in the site, the distance between tooltip and the target is 8px, so top value should be 2rem.
Try to use sdds spacing element css variable for all spacing values. If it is not available in the spacing tokens, then you can fill in your own value. **var(--sdds-spacing-element-xxx)**

components/src/components/tooltips/tooltip.scss
```diff
@@ -0,0 +1,99 @@
...
+  left: 32rem;
```
helloimela, line 72: left position will vary depends on the position of the tooltip. Might be set from the script not from the css, so, maybe add a 0 value in the css, to normalize its position

niklasros:
The tooltip is currently behind a button by default, so I can not preview it.

tooltip

iths-jonathan-engman:
Todo:
--Transition(s)
--Fix small issues in positioning
--Remove border option in Storybook

package.json
```diff
@@ -34,5 +34,8 @@
       "Improvement": ":seedling: Improvements",
       "bug": ":bug: Bug Fix"
     }
+  },
+  "dependencies": {
+    "@popperjs/core": "^2.9.1"
```
helloimela, line 39: popper.js is a dependency to components package, should be installed inside components/ folder. Dependency should be added in sdds/components/package.json

helloimela (changes requested):

components/src/components/tooltips/tooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+            offset: [0, 8],
+          },
+        },
+      ],
+    });
+
+    this.target.addEventListener('mouseenter', () => {
+      this.show = true;
+      //TODO fix improve performance in the future, destroy poppers when it's not visible.
+      // popperInstance.setOptions({
```
mmexvr, line 69: Remove unused code

package.json
```diff
@@ -34,5 +34,6 @@
       "Improvement": ":seedling: Improvements",
       "bug": ":bug: Bug Fix"
     }
-  }
+  },
+  "dependencies": {}
```
helloimela, line 38: Remove unused properties from code changes

helloimela (changes requested):

helloimela (approved):

niklasros:
Do we need all these types of positions in the storybook example? The idea of the different styles of the tooltip is that a specific style should be used depending on where the component is placed. Is it placed or instance in a top left corner, we show this by having the hard corner on the left side of the tooltip.

I've tried to explain in the image below how I see the different usage needs, but I might be missing something!

tooltip-positions

helloimela:
> Do we need all these types of positions in the storybook example? The idea of the different styles of the tooltip is that a specific style should be used depending on where the component is placed. Is it placed or instance in a top left corner, we show this by having the hard corner on the left side of the tooltip.
> 
> I've tried to explain in the image below how I see the different usage needs, but I might be missing something!
> 
> tooltip-positions

@niklasros Yes, from a development perspective, we need to define a speicific value to make the component understand which corner should be border-radius=0 when in a specific position.

niklasros:
> > Do we need all these types of positions in the storybook example? The idea of the different styles of the tooltip is that a specific style should be used depending on where the component is placed. Is it placed or instance in a top left corner, we show this by having the hard corner on the left side of the tooltip.
> > I've tried to explain in the image below how I see the different usage needs, but I might be missing something!
> > tooltip-positions
> 
> @niklasros Yes, from a development perspective, we need to define a speicific value to make the component understand which corner should be border-radius=0 when in a specific position.

Ok! Still, some of the visualizations for me are still wrong in a sense. For instance, when using the bottom-right-corner the tooltip shouldn't be placed like that. But if it make sense from a development point of view, lets go with it!

tooltip_example

niklasros (approved):

PR #87, opened 2021-03-03
AB#474 Feature/dropdown filter

helloimela:
**Describe pull-request**  
- Added feature "Filter" on dropdown
- Added as a separate component `sdds-dropdown-filter`
- Add the options via `data` attribute which should be an array of objects (See storybook to see how use it in code)

**Solving issue**  
Fixes: [AB#474](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/474)

**How to test**  
- Click on storybook > Dropdown > Filter
- See Figma for UI
- Check if filtering works
- Might want to check if it is used in another framework (Angular or React example page)'

**Additional notes**

Merge #85 first

### DIFF
--- a/components/src/components/dropdown/dropdown-option.tsx
+++ b/components/src/components/dropdown/dropdown-option.tsx
@@ -13,7 +13,8 @@ import {
     @Element() host: HTMLElement;
 
     @State() selected:boolean=false;
-    @State() optionLabel:string;
+    // Used as a fallback if value prop is not recognized to match handleClick
+    @State() innerValue:string;
 
     /** Value is a unique string that will be used for application logic */
     @Prop() value:string;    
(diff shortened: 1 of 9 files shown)


niklasros:
It looks great now. My only comment is that on "Label inside" the placeholder text is set as _Select option_ and label text as _Label text_ as default, which makes it looks like two different texts instead of one (_as the text field_) when the user has done a selection. 

I understand that the text is editable in Storybook for instance, so maybe it's not an issue. It just looks a tad off when interacting with the component in Storbyook when its default and you haven't edited the texts :)

niklasros (approved):

mmexvr:
Minor thing with the dropdown filter.

Maybe make the whole element target for writing, not just the input field it self, so a user can click on the whole thing for writing

mmexvr (approved):

PR #110, opened 2021-03-23
Improvement/update demo html

iths-jonathan-engman:
**Describe pull-request**  
Showcasing components with a html demo

**How to test**  
Run live server on demo/html/index.html

**Additional context**  
Feel free to comment on the design and text content


mmexvr (changes requested):
Update the readme file in the html project also, that is the most important one 

You can rewrite the whole thing

Maybe something like this can be added: 
- Explain what it is
- How to install theme and components

demo/HTML/html.css
```diff
@@ -0,0 +1,94 @@
+@import '~node_modules/@scania/grid/dist/scss/grid';
+
+@media (min-width: 1312px) {
+  .sdds-col-xxlg-16 {
+    padding: 0;
```
mmexvr, line 5: We shoudn't effect the grid css at all, since that is what people use to position things. If we effect the grid from our own css that we can make sure the layout works

demo/HTML/html.css
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+body {
+  background: #f6f6f7;
```
mmexvr, line 10: Try and use our variables, maybe add the `var(--sdds-grey-100)` or something, have a look at our background color on SDDS website

demo/HTML/html.css
```diff
@@ -0,0 +1,94 @@
...
+}
+
+p {
+  font-size: 16px;
```
mmexvr, line 14: Same here, try and use the typography package to set the size of the p element. Follow the typography package documentation

demo/HTML/html.css
```diff
@@ -0,0 +1,94 @@
...
+  font-size: 8rem;
+  line-height: 8rem;
+  letter-spacing: -0.04em;
+  padding-bottom: 10rem;
+  height: 48rem;
+  margin-top: 0;
+}
+
+.header-greeting {
+  color: #9CA5C2;
```
mmexvr, line 65: Use our css variables instead

demo/HTML/html.css
mmexvr: Or use classes that are avaliable

mmexvr (changes requested):
Try and use our variables when setting size/color and so on from our foundation :)

demo/HTML/html.css
mmexvr: or use classes on the p to make it work

helloimela (approved):
Nice! good job!

mmexvr (approved):

PR #146, opened 2021-05-19
Feature - Banner component styling

mmexvr:
**Describe pull-request**  
- New component banner
- Removed some unused code

**Solving issue**  
Fixes: [AB#831](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/831)

**How to test**  
Visit storybook either locally or with this preview link below
- Try different states
- Try different options, without prefix, subheadline and so on
- Screen size

**Additional context**  

Demo has c-icon component to show a icon

### DIFF
--- a/components/src/components/banner/banner.scss
+++ b/components/src/components/banner/banner.scss
@@ -19,12 +19,20 @@
 }
 
 .#{$prefix}-banner-prefix {
+  display: inline-flex;
   flex-basis: 16px;
-  width: 16px;
   height: 16px;
   width: 100%;
   margin-right: var(--sdds-spacing-element-12);
   font-size: 4rem;
+  color: var(--sdds-banner-prefix-default-color);
+
+  .#{$prefix}-banner-info &  {
+    color: var(--sdds-banner-prefix-info-color);
+  }
+  .#{$prefix}-banner-error & {
+    color: var(--sdds-banner-prefix-error-color);
+  }
 }
 
 .#{$prefix}-banner-body {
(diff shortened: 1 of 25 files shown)


vichtertz (changes requested):
- Could you put a first capital letter by default on the Subheader text? 
Screenshot 2021-05-19 at 15 52 21

- The icon should be filled icon 
Screenshot 2021-05-19 at 16 08 54

- Could you change the name in the Controls:
  Headline to Header
  Subheadline to Subheader 
(in order to use the same terminology as in the documentation)

Screenshot 2021-05-19 at 16 11 11

mmexvr:
> * Could you put a first capital letter by default on the Subheader text?
>   Screenshot 2021-05-19 at 15 52 21
> * The icon should be filled icon
>   Screenshot 2021-05-19 at 16 08 54
> * Could you change the name in the Controls:
>   Headline to Header
>   Subheadline to Subheader
>   (in order to use the same terminology as in the documentation)
> 
> Screenshot 2021-05-19 at 16 11 11

1. That is up to the user, but I will change in the demo that it is uppercase. I was just lazy :) 
3. I will change the naming for the controls and classes, agree it should match

vichtertz (approved):

components/src/components/banner/banner.scss
```diff
@@ -0,0 +1,62 @@
...
+  &.#{$prefix}-banner-info {
+    background-color: var(--sdds-banner-bg-info);
+  }
+
+  &.#{$prefix}-banner-error {
+    background-color: var(--sdds-banner-bg-error);
+  }
+}
+
+.#{$prefix}-banner-prefix {
```
helloimela, line 21: In the future, this will contains icon, I assume the colour of the icons will follow the state. Maybe we can add the color to the prefix?
So the children will always inherit colour from this classes, and it will be flexible what is inside. In the demo we can use fill=currentColor in the rect svg. And if people want to add their own icon such as <i class=fa fa-book"></i>, it will also follow the colour from banner-prefix.

mmexvr: Changed!

helloimela (approved):
Great work!

helloimela (approved):

PR #179, opened 2021-08-18
Added feature/slider-draft

kjec2f:
**Describe pull-request**  
  Added slider-draft
**Solving issue**  
Fixes: #

**How to test**  
1. Go to storybook > component > slider
2. Desgin similar to "https://www.figma.com/proto/x0Lak62QgvbOuXW8bZFb9D/slider?page-id=80%3A379&node-id=80%3A578&viewport=677%2C691%2C0.11236094683408737&scaling=min-zoom"

**Screenshots**  

**Additional context**  


mmexvr:
conflict

mmexvr:
I think there is an important prop missing, so if you have a range it should be possible to change the value it should be between. So if you have a price range for a product you might need to have bigger or smaller numbers.

So it should be possible to change the `min` and `max` attribute on the html to your need, because it might be needed for developers to change that attribute and use it. You might have a range between 2000-3000 and not 0-100 all the time.

Same with `value` attribute for the input element is important that they can access and use, they it should be able to set a default start value or not

`<input type="range" min="1" max="100" value="50" class="slider" id="myRange">`

components/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,225 @@
...
+        ></input>
+        <div
+          class="range-slider"
+          style={{
+            '--minval': `${this.rangeStyle['--val']}`,
+            '--maxval': `${this.secondRangeStyle['--val']}`
+          }}
+        >
+          <input
+            min="1"
```
mmexvr, line 185: Should be able to customise this value

mmexvr: Same with the other min and max on other examples

components/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,225 @@
...
+          ></input>
+        </div>
+        <input
+          min="1"
+          max="100"
+          type="text"
+          value={`${this.secondRangeStyle['--val']}`}
+          ref={(el) => (this.rightInputTextRef = el as HTMLInputElement)}
+          class="input-text"
+          onKeyUp={(ev) => this.handleOnChangeRight(ev)}
```
mmexvr, line 210: Can we use just on change and not keyup, or is there a specific reason for keyup?

components/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,225 @@
...
+  leftInputTextRef!: HTMLInputElement;
+  rightInputTextRef!: HTMLInputElement;
+
+  @State() leftRangeOldValue: string;
+  @State() rightRangeOldValue: string;
+
+  @State() rangeStyle = {
+    '--min': '1',
+    '--max': '100',
+    '--val': '20'
```
mmexvr, line 25: value should be default 0 and be customisable for the user, what the default value should start on

mmexvr (changes requested):
Above comments :)

alexlioliopoulos:
We are also missing the hover state in the component.

mmexvr:
Can you seperate the examples so it has its own page in storybook, so duel point has one menu for example

components/src/components/slider/slider.tsx
kjec2f: In order to detect the value and show error if the value is not correct before the user release

alexlioliopoulos:
From my point of view, great work!

mmexvr:
@alexlioliopoulos Should the slider be the size of where it is placed or is it always the same small size, can it be for example 6 col big?

alexlioliopoulos:
> @alexlioliopoulos Should the slider be the size of where it is placed or is it always the same size, 100px in this case, or can it be for example 6 col big instead?
> 
> I would assume there are cases where you want a bigger slider?

Great point! I suppose you are right, so we should give devs the option to change the width of it (but not the height).

mmexvr:
> > @alexlioliopoulos Should the slider be the size of where it is placed or is it always the same size, 100px in this case, or can it be for example 6 col big instead?
> > I would assume there are cases where you want a bigger slider?
> 
> Great point! I suppose you are right, so we should give devs the option to change the width of it (but not the height).

Yes! So instead of having a set width on the slider, remove that and let it grow with the parent element instead. So if the parent is 200px width, then the slider will be that.

mmexvr (approved):

mmexvr:
Looks good! 

Some options to add in the future:
- Add step option
- How the state should be when they are above each other

PR #197, opened 2021-09-20
Feature/data table

kjec2f:
**Describe pull-request**  
Adding Data-table base implementation 

**Solving issue**  
Fixes: 

**How to test**  
1. Go to storybook > Table

**Screenshots**  

**Additional context**  


alexlioliopoulos (changes requested):
Great!

My only comments are the following:

- The table header is not as it should be, but I know you already know that.

- When hovering over a cell, the whole row is highlighted. (my bad! i didn't clarify this properly)
image

- When hovering over a column header, the whole column needs to be highlighed. (first time I say that too! sorry)
image

And a question! When adding a super long text in a cell, I know that we discussed having the product team deciding if they want to have a line break or not. But I was wondering, can they actually do that?

alexlioliopoulos:
Looks good!
One small change that I would suggest is the following:
I think we should not have curved corner in the column header, until we fix the problem with the caption (table header).
image
It now feels like it's buggy, but if we could make it have an edge corner, it would feel better.
But at the same time, when no caption (table header) is used, then we want it to have a curved corner.

alexlioliopoulos (approved):
Looks good! The column hover is going to be left out for now, right?

PR #225, opened 2021-11-04
Open modal with multiple buttons

Johnny-Scania:
**Describe pull-request**  
Enabling modal component to be opened from multiple buttons

**Solving issue**  
Fixes: #[1093](https://dev.azure.com/scaniadigitaldesignsystem/SDDS/_workitems/edit/1093)

**How to test**  
1. Run Storybook
2. Open Modal component
3. Try to open same modal (modal one) from two different button - primary and secondary

**Additional context**  
- In Documentation: Advise using classes instead of ids in HTML 
- check the use of modal-dismiss (is it OK to use it in button)

### DIFF
--- a/components/src/components/modal/modal.tsx
+++ b/components/src/components/modal/modal.tsx
@@ -8,6 +8,7 @@ import {
   Element,
   Watch,
 } from '@stencil/core';
+
 @Component({
   tag: 'sdds-modal',
   styleUrl: 'modal.scss',
@@ -16,8 +17,10 @@ import {
 export class Modal {
   // Target selector to show modal
   @Prop() selector;
+
   // Disable click event on backdrop
   @Prop() prevent = false;
+
   // Size of the modal
   @Prop() size = 'md';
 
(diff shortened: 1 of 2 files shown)


components/src/components/modal/modal.tsx
```diff
@@ -47,7 +47,7 @@ export class Modal {
   }
 
   dismissModal() {
-    const nodes = this.el.querySelectorAll('[modal-dismiss]');
+    const nodes = this.el.querySelectorAll('[data-dismiss-modal]');
```
Johnny-Scania, line 50: @mmexvr  and @gaonita - I think we can think of something like this for naming convention, for example, if we need to close another component in a similar way, let say toaster, it could be called "data-dismiss-toaster".

gaonita (approved):
it works nicely!

mmexvr:
Looks good! Just testing the last stuff

Would be good if all examples got updated also. Everything under demo is updated

mmexvr (approved):

PR #229, opened 2021-11-08
Utility classes

Johnny-Scania:
**Describe pull-request**  
Making a basic set of utility classes

**Solving issue**  
Fixes: #[AB1151](https://dev.azure.com/scaniadigitaldesignsystem/SDDS/_workitems/edit/1151)

**How to test**  
1. Open utility classes file
2. Run local environment
3. Try to add one of the utility classes to a HTML component

### DIFF
--- a/components/src/components/theme/theme.scss
+++ b/components/src/components/theme/theme.scss
@@ -21,7 +21,7 @@
 @import '../../patterns/footer/footer.scss';
 
 // Utility classes
-@import "../../helpers/_utility-classes.scss";
+@import '../../helpers/_utility-classes.scss';
 
 html.#{$prefix},
 body.#{$prefix},
--- a/components/src/components/theme/theme.scss
+++ b/components/src/components/theme/theme.scss
@@ -20,6 +20,9 @@
 @import '../../patterns/side-menu/side-menu.scss';
 @import '../../patterns/footer/footer.scss';
 
+// Utility classes
+@import "../../helpers/_utility-classes.scss";
+
 html.#{$prefix},
 body.#{$prefix},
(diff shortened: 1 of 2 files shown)


mmexvr:
One thing that I think needs to be added is` !important`, since these rules need to be applied so they don't fall behind an inline rule or something. What do you think @Johnny-Scania ? since stuff like ID/inline will overrule these rules

Johnny-Scania:
> One thing that I think needs to be added is` !important`, since these rules need to be applied so they don't fall behind an inline rule or something. What do you think @Johnny-Scania ? since stuff like ID/inline will overrule these rules

That is a good idea. I will add it ;)

mmexvr:
Looks good! Great idea adding this! I think this is missed a lot. Just some small questions/opinions

1. I was thinking about the `-u-`, I would assume it stands for utility. I don't see the need for that part besides making it more unique. I think `.#{$prefix}-relative` is fine or maybe add `.#{$prefix}-pos-relative` or  another example`.#{$prefix}-pad-p0/ .#{$prefix}-p-p0 ` instead of `.#{$prefix}-u-p0`, if we need to make it more unique for us. 

2. I think this is a really good start what is included, later on we can add shadows,float, text alignment https://getbootstrap.com/docs/5.1/utilities/display/

3. Maybe we should add mixins making it easier to add more options
```$utilities: (
  "opacity": (
    property: opacity,
    values: (
      0: 0,
      25: .25,
      50: .5,
      75: .75,
      100: 1,
    )
  )
);
```

mmexvr (approved):

gaonita (approved):

PR #232, opened 2021-11-11
Improvement/side menu in React demo

gaonita:
**Describe pull-request**  
Added side menu example instead of header menu in React demo

**Solving issue**  
Fixes: #AB1154

**How to test**  
1. Run React demo 
2. Check in side menu 
     - Dropdown menu items should open sub-menu items 
     - Responsive(Mobile) view: The menu is always present on the left side of the header in these viewports.
     - Toggle between the expanded and collapsed state by pressing the expand/collapse link in the bottom of the side menu and the pop-over appears at hover of sub-menu item

**Known issue**
- Focus state(outline) on menu items don't cover the whole item box area. It is related to shared classes with header
- Active status example is missing


Johnny-Scania:
This is an issue I get when starting my local. I will try to find an issue causing it.
bug_react

Johnny-Scania:
@mmexvr and @alexlioliopoulos: Question since you might know more about expected behaviour.
Take a look at this video:

https://user-images.githubusercontent.com/91877950/141752469-88492759-61e9-4cd5-a322-26689e3c1e99.mp4

So, the Side menu is as long as the page is, which makes closing it a bit tricky here. I am guessing that the height of the side menu should same as the "window" height and not page height, but better double-check. Depending on what is expected behaviour we can see if the component itself is missing some classes or it needs to be wrapped with something in React to display properly.

gaonita:
> This is an issue I get when starting my local. I will try to find an issue causing it. bug_react
oh that's because of my poor naming for the file. the lowercase and uppercase S for the 'side menu' got mixed up. will try to clean it up!

gaonita:
> @mmexvr and @alexlioliopoulos: Question since you might know more about expected behaviour. Take a look at this video:
> 
>  Extra_long_scroll_collapse_sidemenu.mp4 
> So, the Side menu is as long as the page is, which makes closing it a bit tricky here. I am guessing that the height of the side menu should same as the "window" height and not page height, but better double-check. Depending on what is expected behaviour we can see if the component itself is missing some classes or it needs to be wrapped with something in React to display properly.

As fas as I understood with existing documentation and code, this was the 'expected behaviour' with what we had for now. but yes! I and Mena thought it is questionable 🤔  so I was going to discuss about the placement of the collapse button in the 'devs meet design' meeting.

Johnny-Scania:
> > @mmexvr and @alexlioliopoulos: Question since you might know more about expected behaviour. Take a look at this video:
> > Extra_long_scroll_collapse_sidemenu.mp4
> > So, the Side menu is as long as the page is, which makes closing it a bit tricky here. I am guessing that the height of the side menu should same as the "window" height and not page height, but better double-check. Depending on what is expected behaviour we can see if the component itself is missing some classes or it needs to be wrapped with something in React to display properly.
> 
> As fas as I understood with existing documentation and code, this was the 'expected behaviour' with what we had for now. but yes! I and Mena thought it is questionable 🤔 so I was going to discuss about the placement of the collapse button in the 'devs meet design' meeting.

Perfect! I think the integration part you did should be part of this release, while any other fix/improvement on the Side menu component should be a separate ticket. Nice work BTW ;)

gaonita:
> > > @mmexvr and @alexlioliopoulos: Question since you might know more about expected behaviour. Take a look at this video:
> > > Extra_long_scroll_collapse_sidemenu.mp4
> > > So, the Side menu is as long as the page is, which makes closing it a bit tricky here. I am guessing that the height of the side menu should same as the "window" height and not page height, but better double-check. Depending on what is expected behaviour we can see if the component itself is missing some classes or it needs to be wrapped with something in React to display properly.
> > 
> > 
> > As fas as I understood with existing documentation and code, this was the 'expected behaviour' with what we had for now. but yes! I and Mena thought it is questionable 🤔 so I was going to discuss about the placement of the collapse button in the 'devs meet design' meeting.
> 
> Perfect! I think the integration part you did should be part of this release, while any other fix/improvement on the Side menu component should be a separate ticket. Nice work BTW ;)

Thanks! yes that's how I wanted to do as well :) thank you for the thorough review even with a recording @Johnny-Scania ⭐

alexlioliopoulos:
Just to clarify that the "Collapse" and "Expand" links should always be sticky.

alexlioliopoulos:
So to elaborate further, the expand and collapse links should always be shown to the user. They shouldn't need to scroll down to collapse, in case many navigation links exist. In a sense, this area should be above the other items in the side menu.

alexlioliopoulos:
But I agree, documentation is not clear about this!

gaonita:
> So to elaborate further, the expand and collapse links should always be shown to the user. They shouldn't need to scroll down to collapse, in case many navigation links exist. In a sense, this area should be above the other items in the side menu.

Thank you for the clarification Alex! Ok then we need to make the collapse button class to be sticky one. I created an task for it in Azure.

Johnny-Scania (approved):
Works and looks good to me!
I propose "Squash and merge" option so it comes down to one commit.
PS: If you have not, make sure you have the latest master in your code ;)

PR #236, opened 2021-11-16
Scale up for second level values of utility classes

Johnny-Scania:
**Describe pull-request**  
Scaling up the second level of utility classes for padding and margin

**Solving issue**  
By mistake values for second-level remaind on value of 8px instead of 16px.


gaonita (approved):

Johnny-Scania:
Since this is a tiny one and already reviewed, I will merge it. Regarding the Utility class display in Storybook, I will make a separate branch for it.

Issue #187, opened 2021-08-26
Bug - Undeline for links in Cards


alexlioliopoulos:
Links inside a card are not supposed to be underlined. Could we make that happen?

Current:
![image](https://user-images.githubusercontent.com/76053160/130956162-6ba6c08c-37a5-4d99-89aa-4373a8b17f5d.png)

Correct:
![image](https://user-images.githubusercontent.com/76053160/130956200-d120fbb0-4611-4d3d-8ce9-06fcad403bd1.png)

[AB#982](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/982)

PR #250, opened 2021-12-09
feat(link): add class for controling underline

theJohnnyMe:
**Describe pull-request**  
Introducing CSS class for controlling link underline appearance.

**Solving issue**  
Fixes: [AB#982](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/982)  #187 

**How to test**  
1. Go to Strotybook
2. Open Card component, select Link variation
3. Links in the Card component should not be underlined anymore
4. Now open Link component, select a Basic variation
5. In Storybook controls it should be "noUnderline" knob to test component with/without Underline

### DIFF
--- a/components/src/components/card/card.stories.js
+++ b/components/src/components/card/card.stories.js
@@ -94,7 +94,8 @@ link.argTypes = {};
 link.args = {
   divider: true,
   text: 'This is a short and consist detail text describing for the user what this text is really about.',
-  footer: '<a href="#">Link text</a><a href="#">Link text</a>',
+  footer:
+    '<a class="sdds-link sdds-link--no-underline" href="#">Link text</a><a class="sdds-link sdds-link--no-underline" href="#">Link text</a>',
 };
 
 export const button = CardTemplate.bind({});
--- a/components/src/components/link/link.scss
+++ b/components/src/components/link/link.scss
@@ -32,4 +32,8 @@ a,
     color: var(--sdds-link-disabled);
     pointer-events: none;
   }
+
+  &--no-underline {
+    text-decoration: none;
+  }
 }
--- a/components/src/components/link/link.scss
+++ b/components/src/components/link/link.scss
@@ -10,15 +10,18 @@ a,
   &:active {
     color: var(--sdds-link);
     text-decoration: var(--sdds-link-text-decoration);
+    text-decoration-color: var(--sdds-link);
   }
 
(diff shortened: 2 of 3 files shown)


alexlioliopoulos (changes requested):
Nice! 
One change that I'd like to see, in order to align with design, is for the no-underline version of our link component, to not have underline on hover. You can check the different states (default, hover, disabled) of the link component inside cards in Figma. Or give me a call!

theJohnnyMe:
> Nice! One change that I'd like to see, in order to align with design, is for the no-underline version of our link component, to not have underline on hover. You can check the different states (default, hover, disabled) of the link component inside cards in Figma. Or give me a call!

Thanks for pointing it out Alex, I forgot about it. Fixing it right away ;)

theJohnnyMe:
I have added a rule for not having an underline on hover for the "no-underline" version of the Link.
Change of colors between regular and hover state on small font and when link is not part of text is hard to notice so here is Photoshop enlargement: 

photoshop_comparisson

mmexvr:
**Looks good!** 


_Just some random thoughts_ 💭 

The naming. In this case `sdds-link--no-underline`, If we should try and keep it shorter. I have been think about this more and more for our components or if it is fine with this. In link component I don't think it is that big of a deal, but header/side-menu that contains such a complex structure it might be hard 😆  I guess this should be a part of our rule set developing a component. 

Another thing that i thought of, if this should be either a utility or a component specific option. I guess depending on the hover state, in this case, the component option would be more fitting.

mmexvr (approved):

theJohnnyMe:
> **Looks good!**
> 
> _Just some random thoughts_ 💭
> 
> The naming. In this case `sdds-link--no-underline`, If we should try and keep it shorter. I have been think about this more and more for our components or if it is fine with this. In link component I don't think it is that big of a deal, but header/side-menu that contains such a complex structure it might be hard 😆 I guess this should be a part of our rule set developing a component.
> 
> Another thing that i thought of, if this should be either a utility or a component specific option. I guess depending on the hover state, in this case, the component option would be more fitting.

I agree with you, I had similar worries. Long time ago I used this one: https://en.bem.info/methodology/css/
We can see if all of it or parts of it can be applied as our rules for CSS development.

Issue #111, opened 2021-03-24
Improvement - Dev enviroment for core folder doesn't have watchers 


mmexvr:
**Describe current situation**  
_Describe what the current situation and what problem it is causing for this improvement_
Under the core folder for foundation, we can't live edit files and see the effected changes directly in the browser.
 We need to rebuild to see the change in core/** 

Theme is running on gulp that have watchers for the theme/light for example

**Describe the improvement**  
_A clear and concise description of what kind of improvement you have in mind._
Somehow make it possbile to rebuild everything under the core/** with storybook running with lerna

**If applicable, please provide steps to reproduce**
_Steps to reproduce the behavior:_  
1. Run storybook
2. Do changes in a core/**/*.scss file
3. Save it
4. Since there is no watcher nothing will happened

[AB#696](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/696)

mmexvr:
It worked before because everything was under the theme/light folder that uses gulp watchers

PR #251, opened 2021-12-09
fix(starting-sdds): bug running sdds the first time

mmexvr:
Running sdds first time right after install won't work since the core packages hasn't been built,
this makes it run build right after install under prepare phase

**Describe pull-request**  
If you clone a fresh sdds repo and run
- Npm i 
- npm start

It won't start, on master, because the theme will fail, reason for this is because the core packages hasn't been built yet. The core package don't have watchers, so every changes in those core packages needs to be built. Lerna start only starts the npm start command, which those packages doesn't have. Issue related with this #111 

**Solving issue**  
This solves the issue so you don't have to run build right before you start up sdds, so this will make it run build during prepare phase in npm

**How to test**  
1. Clone new
2. npm i  (during prepare with husky, build will run also)
3. npm start

Try the same commands on master

**Additional context**  
#111  Related issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "js-lint-fix": "eslint . --ext .js,.jsx,.ts,.tsx --fix",
     "lint-check": "npx stylelint **/*.scss && eslint . --ext .js,.jsx,.ts,.tsx && prettier --check .",
     "lint-all": "npx stylelint **/*.scss && eslint . --ext .js,.jsx,.ts,.tsx && prettier --write .",
-    "prepare": "husky install",
+    "prepare": "husky install && npm run build",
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
Tested two fresh cloned repos. One running your branch, one running master. Yours one runs with no issues, master one reported missing typography and storybook could not resolve custom-elements.json and collection/index files so it failed.  

Your change looks good to me. Good idea to have it in prepare script so it runs smoothly. nice idea Mena ;)

theJohnnyMe:
Only see if you can rebase locally, push and then merge to keep the git log clear...

PR #271, opened 2022-01-05
refactor(header): decoupling and simplifying code

theJohnnyMe:
**Describe pull-request**  
Refactoring Header component

**Solving issue**  
Fixes: [AB#1149](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1149)

**How to test**  
1. Go to Storybook
2. Open Header component and test if all features work as intended
3. Open Demo examples and test header component


components/src/components/theme/theme.scss
```diff
@@ -23,6 +23,11 @@
...
 @import '../../helpers/_utility-classes.scss';
 
+* :focus {
+  outline: 2px solid var(--sdds-blue-400);
+  outline-offset: -2px;
+}
+
 html.#{$prefix},
 body.#{$prefix},
 .#{$prefix}-body {
```
theJohnnyMe, line 33: Can we distrubute these rule in component related files as some of them are related to specific component and other one in newly created files like sdds-reset.scss OR other common/shared sccs file?

components/src/patterns/side-menu/side-menu.scss
```diff
@@ -2,8 +2,11 @@
 @import '@scania/typography/dist/scss/mixins';
 @import '@scania/typography/dist/scss/tokens';
 @import '@scania/grid/dist/scss/grid.scss';
-@import '../header/mixins';
-@import '../header/navbar-collapsible';
+@import 'mixins';
```
theJohnnyMe, line 5: Until refactoring of side-menu, let these files remain here as side-menu consumes much more classes then mixins and navbar-collapsible.

components/src/patterns/side-menu/side-menu.stories.js
```diff
@@ -31,10 +31,9 @@ const Template = (args) => {
     .sb-show-main.sb-main-padded {
       padding: 0;
     }
-    .sdds-demo-container {
-      padding-top: var(--sdds-header-height);
+    .sdds-demo-container { 
       align-items: stretch;
-      height: 100%;
+      height: calc(100% - 64px);
```
theJohnnyMe, line 36: Depends if header is fixed or scrolls with the page. It up to user to define.

components/src/components/theme/theme.scss
```diff
@@ -23,6 +23,11 @@
 // Utility classes
 @import '../../helpers/_utility-classes.scss';
 
+* :focus {
+  outline: 2px solid var(--sdds-blue-400);
```
mmexvr, line 27: Nothing to be acted on right now, but interesting to look into

We had a chat about this, just open question. If we should move this to components.scss in component package. Might be a good idea. 

Personal note, I would like to keep theme.scss just to have imports. Or in the future if the theme-component will even be needed

mmexvr:
💡  Idea

Should we move everything header related to be under components instead of patterns? 
The idea before was that the header was more of a pattern but we split it up to be a component instead

components/src/components/theme/theme.scss
theJohnnyMe: Definitely I agree, in next works around component I will see if I can find or come up with a file that will contain different rules applicable to all components. Something that will contain reset rules and sdds common rules.

theJohnnyMe:
> 💡 Idea
> 
> Should we move everything header (and side-menu) related to be under components instead of patterns? The idea before was that the header was more of a pattern but we split it up to be a component instead

I think that could also be a good idea as in website we keep under components too. It is a very complex one, but still, it is a component.

theJohnnyMe:
> Super satisfied to see this refactored code! Thanks for doing this 💚 I think you are already going to but please make sure side menu works fine too when you merge this to master which I didn't last time 🙈

Valid point! I have tested it within Angular and React where Side-menu exists too - it looks promissing.

PR #286, opened 2022-01-20
style(dropdown): restyling dropdown to fit along with other inputs

gaonita:
**Describe pull-request**  
Restyle dropdown to fit along with other inputs
- No border radius on bottom edges
- Border only on the bottom 
- Border color on different states
  - default state: change to grey400
  - hover state: change to grey600
  - error state: change to semantic negative
  - focus state: change to blue400 and width 2px
  - active state: change to grey800
  - disabled state: no border 

Figma: https://www.figma.com/file/jZONFYFlckRROjSVPPu2DD/Dropdown-exploration?node-id=2%3A163

**Solving issue**  
Fixes: [AB#1114](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1114)

**How to test**  
1. Go to storybook dropdown example
2. Check the look of dropdown in different state

### DIFF
--- a/components/src/components/dropdown/_dropdown-core.scss
+++ b/components/src/components/dropdown/_dropdown-core.scss
@@ -1,31 +1,37 @@
 @mixin dropdown-wrapper {
   @include type-style('detail-02');
+
   color: var(--sdds-dropdown-text);
   display: flex;
   align-items: center;
   background-color: var(--sdds-dropdown-bg);
   cursor: pointer;
-  border: 1px solid var(--sdds-dropdown-border);
   transition: box-shadow 0.1s ease-in;
-  border-radius: var(--sdds-dropdown-border-radius);
+  border-radius: 1rem 1rem 0 0;
+  border: 0;
+  border-bottom: 1px solid var(--sdds-dropdown-border-bottom);
   box-shadow: 0;
   width: 100%;
   text-align: left;
   outline: none;
-  padding: var(--#{$prefix}-dropdown-padding-top-large)
+  padding:
+ var(--#{$prefix}-dropdown-padding-top-large)
     var(--#{$prefix}-dropdown-padding-left);
 
   &:hover {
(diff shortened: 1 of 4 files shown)


alexlioliopoulos (changes requested):
Super good, nice job Gaon!
But since I need to find a reason why Scania is paying my salary, I have to comment on one small thing. 

I think that you are using grey-958 for both the "Select option" text, meaning the text that appears before user makes any selection, and for the options themselves.

According to design, the correct value is grey-700 for the "Choose an option" text, grey-900 for all the options, and grey-958 for the Option that the user has selected and is shown in the top part of the dropdown.

mmexvr:
on @alexlioliopoulos  comment 

- placeholder text is grey 700. Since it is a button it doens't have placeholder like textfield it will take the regular color.  Maybe have a class when a option is not picked that is using the placeholder color 
image

gaonita:
> 
this is 

> Another thing I noticed after having a look at the other variants:
> 
> The "label-inside" variant should be like this:
> 
> image
> 
> Now it's like this:
> 
> image
> 
> Meaning that in the current version, there is no label present, but we keep showing the "select option" text.

I see! then I will create a new ticket for this issue since it looks like more than restyling.

gaonita:
> Super good, nice job Gaon! But since I need to find a reason why Scania is paying my salary, I have to comment on one small thing.
> 
> I think that you are using grey-958 for both the "Select option" text, meaning the text that appears before user makes any selection, and for the options themselves.
> 
> According to design, the correct value is grey-700 for the "Choose an option" text, grey-900 for all the options, and grey-958 for the Option that the user has selected and is shown in the top part of the dropdown.

Fixed! please check it out again @alexlioliopoulos thank you for the tip @mmexvr !

alexlioliopoulos (approved):
Oh yes!

mmexvr (approved):

PR #287, opened 2022-01-21
test(test): test

gaonita:
test aws

**Describe pull-request**  

**Solving issue**  
Fixes: #

**How to test**  
1. Go to...
2. Check in...
3. Run ...

**Screenshots**  

**Additional context**  


theJohnnyMe:
@gaonita  - Did you just make Storybook spins on AWS? If so, awesome, really awesome work!

gaonita:
> it doesn't work perfectly yet🙈  therefore this test PR is messy. but I'm trying! anyway thanks! :D

mmexvr:
Super cool! Would be awesome if this worked. 🚀  ☁️ 

I have two things to look at here :)
- On netlify we don't use start command. We have a separate command called netlify that only builds and uses the dist from storybook, under component folder, for the preview.
- Keep track how the cost work for that service, cost management service in AWS you can se everything

components/.storybook/preview.js
```diff
@@ -1,8 +1,8 @@
 import { setCustomElements, addDecorator } from '@storybook/web-components';
-import { withThemes } from 'storybook-addon-themes/html';
+// import { withThemes } from 'storybook-addon-themes/html';
 import customElements from '../dist/collection/custom-elements.json';
 
-addDecorator(withThemes);
+// addDecorator(withThemes);
```
theJohnnyMe, line 5: I noticed that error is not related to the content, but to the storybook and aws. It felt that some of the code is not aligned with standard, while Netlify does not complain, aws seems to do. Anyway, keep on, it looks like you are on a good way to solve it. I would like to hear your impressions of it tomorrow during check in.

Issue #258, opened 2021-12-20
Bug - Opening dropdown in sidemenu causing a jump


mmexvr:
**Describe the bug**  
_A clear and concise description of what the bug is._
When opening the dropdown in sidemenu, in angular demo, will make the option jump around. See video
This is caused by border-left having 0 instead of being transparent

**To Reproduce**   
_Steps to reproduce the behavior and be very detailed_  
1. angular demo
2. Open a dropdown in sidemenu

**Expected behavior**  
_A clear and concise description of what you expected to happen._
Not the jump around

**Screenshots**  
_If applicable, add screenshots to help explain your problem._

https://user-images.githubusercontent.com/35451568/146754351-89ab25ee-d44b-4110-9da4-f285d37c9d03.mov

**Version of SDDS or which package**  
 _E.g 1.2.3 or @scania/
1.5.2

**System information:**  
_Please complete the following information, don't forget version_
 - Framework: angular 11
 - Device: MacBook
 - OS: MacOS
 - Browser: Chrome

**Additional context**   
_Add any other context about the problem here._

[AB#1312](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1312)

PR #288, opened 2022-01-24
fix(side-menu): side menu dropdown item opened border fix

gaonita:
**Describe pull-request**  
1. Fixed the bug that opening the dropdown in side menu makes the option jump around. This is caused by border-left having 0.
 
2. Fixed Dropdown placeholder color on disabled state

**Solving issue**  
Fixes: #258 
Fixes: [AB#1312](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1312)

**How to test**  
1. Go to Storybook 
2. Check in side menu dropdown item alignment on opened state(you can compare it with current storybook example) 

drawing <- Current alignment     drawing <- Fixed alignment

3. Check in dropdown component placeholder color when it is disabled state it should be screenshot below
image

### DIFF
--- a/components/src/components/dropdown/dropdown.scss
+++ b/components/src/components/dropdown/dropdown.scss
@@ -114,6 +114,9 @@
       pointer-events: none;
       color: var(--sdds-dropdown-text-disabled);
       border: none;
+      .#{$prefix}-dropdown-placeholder {
+        color: var(--sdds-dropdown-text-disabled);
+      }
     }
   }
 }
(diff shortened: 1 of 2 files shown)


gaonita:
> Do I need to verify this using the demo? I can't figure out how to verify this using only storybook since the component isn't interactive (collapsible) there.

@areblk you can use storybook control panel to open/close the dropdown. And yes it will be super helpful if you can test these type of changes in the demos. perhaps you can learn how to create a link to certain commit and check the demos? let's discuss in dev check in.

theJohnnyMe (approved):
Looks OK to me.

areblk (approved):
Comparing this to the master branch I can verify that the jump is fixed. Tested in Edge.

PR #296, opened 2022-01-28
refactor(storybook): clean up unused addDecorator in storybook preview.js

gaonita:
**Describe pull-request**  
Clean up unused import(`addDecorator, withThemes`) to properly use AWS amplify preview
Fixes: [AB#1400](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1400)

**How to test**  
Check the preview from AWS amplify shows the current storybook example preview.


areblk (changes requested):
Not sure if this change is the cause of the problems, but i noticed some weird differences between the amplfify and the netlify. I'd consider the first one a bug.

Screen images compare amplify (to the left) with netlify master (to the right)

**1. Docs-tab looks very different in amplify compared to netlify. The main difference is that code in "preview-boxes" doesn't render a preview, only showing code.**

Docs look different-04
Docs look different-01
Docs look different-02
Docs look different-03

**2. The top part of Canvas-tab is shorter in amplify than in netlify.**

Amplify preview shorter window-01
Amplify preview shorter window-02

theJohnnyMe (changes requested):
Please check package.json in components level and main.js in .storybook - there are some imports left for storybook-addon-themes. Clean it, check if it works. I think after that we are good to go.

theJohnnyMe:
> Not sure if this change is the cause of the problems, but i noticed some weird differences between the amplfify and the netlify. I'd consider the first one a bug.
> 
> Screen images compare amplify (to the left) with netlify master (to the right)
> 
> **1. Docs-tab looks very different in amplify compared to netlify. The main difference is that code in "preview-boxes" doesn't render a preview, only showing code.**
> 
> Docs look different-04 Docs look different-01 Docs look different-02 Docs look different-03
> 
> **2. The top part of Canvas-tab is shorter in amplify than in netlify.**
> 
> Amplify preview shorter window-01 Amplify preview shorter window-02

Good catch @areblk! I have spent some time trying to find a workaround but I could not at the moment. I think there is something in the way we write our story file. Docs tab is not so much used, so far we rely only on Canvas one to make sure our components work OK. If you ask me, I would go with this bug as fully working Amplify is much more valuable to us compared to the Docs tab in Storybook. I checked the version on master branch - not many components have it done in a proper way anyway. Kind of: Is it a issue? - it is. Is it a blocker for us - no.

I would suggest reporting this as an issue in our backlog so we do not forget about it and we get back to it. Once again, nice catch really ;)

gaonita:
> Great catch indeed @areblk! I will try to make a preview using Netlify with this changes and compare them to be sure if this change caused the issue. 
But yes, I agree with you @theJohnnyMe. The docs addon is something we need to improve the way of use later anyway. I will clean up the package.json and let you know again :) 
Thank you for reviewing!

gaonita:
> 
That I checked before and it is without '@'. you can search the addons by names and see how you should import.  
 https://www.npmjs.com/package/storybook-addon-themes

areblk:
Last commit removed the option of changing theme per component. Now it's only possible to change background, but not the theme of the component (`on-white` or `on-grey´). Was this the intention?
Theme-preview-removed

gaonita:
Now I made a preview with netlify and the docs doesn't get properly rendered in there neither :(  might be the addon version is outdated... but like we said let's leave it as it is now and revisit later? I cleaned up the package.json @theJohnnyMe. Great work @areblk and thank you for making a report.

gaonita:
> Last commit removed the option of changing theme per component. Now it's only possible to change background, but not the theme of the component (`on-white` or `on-grey`). Was this the intention? Theme-preview-removed
Oh maybe we are using that addon there! awesome you noticed @areblk! I'm adding it to the package again @theJohnnyMe  :D haha it's never ending task for me

areblk:
Theme preview is back, but it's in another place compared to main branch.

f6d0804-theme-preview

f6d0804-background-preview

components/package.json
```diff
@@ -50,6 +50,7 @@
     "@storybook/addon-links": "^6.2.9",
     "@storybook/addon-notes": "^5.3.21",
     "@storybook/web-components": "^6.2.9",
+    "storybook-addon-themes": "^6.1.0",
```
theJohnnyMe, line 53: Good to have it next to each other, easy to spot all of it.

theJohnnyMe (approved):
There are things we can improve in Storybook, like the removal of the double theme icon, but that one seems to be appearing in old code too and hence is outside of the area of this branch.

Nice exploration and really nice work that simplifies our development. Keep on rocking Gaon! 
Thanks for finding time to work on it.

areblk:
OK with current state

areblk (approved):

PR #305, opened 2022-02-14
fix(storybook): using layout for arranging content in stories

theJohnnyMe:
**Describe pull-request**  
Solving issues with Storybook padding, margins, and centralizing content.

**Solving issue**  
We used tricks to remove padding or margins set by stories.
There is a layout attribute that can be used instead, a very elegant solution for this issue.
More about it: https://storybook.js.org/docs/react/configure/story-layout

**How to test**  
1. Open Storybook
2. All simple component stories need to be centralized
3. Other stories for full-screen components have set fullscreen layout property (header/footer/side menu/colors/....)

### DIFF
--- a/components/.storybook/preview-body.html
+++ b/components/.storybook/preview-body.html
@@ -1,8 +1,0 @@
-<style>
-    .sb-show-main.sb-main-padded {
-        margin: 0;
-        padding: 0;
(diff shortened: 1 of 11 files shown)


gaonita (approved):
Neat!

PR #308, opened 2022-02-16
feat(chromatic):Visual regression testing tool for buttons

areblk:
**Describe pull-request**  
- Adding chromatic as a tool for visual regression testing. This branch is a pilot only enabling this for buttons.
- Chromatic snapshot individual stories, so I've refactored button stories to _represent_ all existing button variations (currently ~900).
- Also trying to add Chromatic as a hook for pull requests.

**Solving issue**  
Fixes: [AB#1335](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1335)

**How to test**  
Not sure, but these steps hopefully open Chromatic
1. Open branch in your local environment
2. Run chromatic from sdds/components with CLI `npm run chromatic`

Try if you can access build via https://www.chromatic.com/manage?appId=620cc1abd4e7e7003aa19d3e&view=collaborate

**Screenshots**  

**Additional context**  


components/src/components/button/button.stories.js
```diff
@@ -1,31 +1,42 @@
...
+  },
+  argTypes: {
     btnType: {
       name: 'type',
       defaultValue: 'primary',
       description:
         'Four different button types to help the user to distinguish the level of importance of the task they represent',
       control: {
-        type: 'select',
+        type: 'radio',
```
theJohnnyMe, line 20: A much better way to deal with variation, really nice fix!
I would say go ahead and change other, similar stories to use radio instead, as a separate ticket of course.

components/src/components/button/button.stories.js
```diff
@@ -1,31 +1,42 @@
...
-  argTypes: {
-    size: {
-      control: {
-        type: 'select',
-        options: ['default', 'small', 'medium'],
-      },
-      defaultValue: 'default',
-      description: 'Size of the button',
+  parameters: {
+    layout: 'padded',
```
theJohnnyMe, line 8: Nice!

components/src/components/button/button.stories.js
```diff
@@ -63,31 +81,31 @@ const ButtonTemplate = ({
       break;
   }
   const fbClass = fullbleed ? 'sdds-btn-fullbleed' : '';
-  const inlineStyle = fullbleed ? 'style="width:200px;"' : '';
-
+  const inlineStyle = fullbleed ? 'style="width:100%;"' : '';
```
theJohnnyMe, line 84: Nice!

components/src/components/button/button.stories.js
```diff
@@ -63,31 +81,31 @@ const ButtonTemplate = ({
...
 
+  // chromatic snapshot requires icon to be sdds-icon instead of font
   return `
   <sdds-theme></sdds-theme>
-  <style>
-    @import url('https://cdn.digitaldesign.scania.com/icons/dist/1.1.0/fonts/css/sdds-icons.css');
-    i {
-      font-size: 4rem;
-    }
-  </style>
```
theJohnnyMe, line 77: Nice that it is removed!

theJohnnyMe:
My question here is can we use Chromatic to pass on different parameters or do you have to create stories that have all settings set before so one story has to be === test case?

Let's find a good way to handle this, otherwise, it would be messy to browse around Storybook with so many substories.


double_values

areblk:
> My question here is can we use Chromatic to pass on different parameters or do you have to create stories that have all settings set before so one story has to be === test case?
> 
> Let's find a good way to handle this, otherwise, it would be messy to browse around Storybook with so many substories.
> 
> double_values

Chromatic can only catch individual stories, not parameters. I agree that number of stories should be kept to a minimum, but enough to represent the range of variants that a button can have so we can verify it. The reason that Chromatic needs a story is that all combinations of parameters quickly reach a great number of test cases.

I've reduced the number of combinations to 12 representative stories. For example, stories for size values are always combined with icons, but this can be changed with individual parameter controls. Do you feel that any of these stories are unnecessary and should be removed?

One way we could do it, that I've seen carbon do with their storybook, is to combine multiple components in one story https://react.carbondesignsystem.com/?path=/story/components-button--expressive-buttons . Perhaps that could reduce number of stories, but still capture visuals with Chromatic. Although, I think I would need help setting that up. Maybe that works better for web components?

alexlioliopoulos:
I am actually not really sure how I can review this

alexlioliopoulos (approved):
I now understand what the problem with paddings and left alignment was. However, it still seems to be that storybook component and reference design are not placed at the same place, so.. is this good enough for us?

components/src/components/button/button.stories.js
```diff
@@ -63,31 +81,31 @@ const ButtonTemplate = ({
...
   <button class="sdds-btn sdds-btn-${btnType} ${sizeValue} ${fbClass} ${
     disabled ? 'disabled' : ''
   } ${onlyIconCss}" ${inlineStyle}>
     <span>${text}</span>
     ${
       icon
-        ? "<span class='sdds-btn-icon'><i class='sdds-icon scania-cross'></i></span>"
+        ? "<span class='sdds-btn-icon'><sdds-icon name='scania-cross'></sdds-icon></span>"
```
gaonita, line 96: yes it's neat!

package.json
```diff
@@ -27,6 +28,7 @@
     "lerna-changelog": "^1.0.1",
     "lint-staged": "^12.1.2",
     "prettier": "2.4.1",
+    "storybook-addon-pseudo-states": "^1.0.0",
```
gaonita, line 31: what does this do? couldn't find where it's used

demo/HTML/button-playground.js
```diff
@@ -0,0 +1,32 @@
+let css = 'sdds-btn sdds-btn-primary';
+let size = ['', 'sdds-btn-md', 'sdds-btn-sm'];
+let fullbleed = ['', 'sdds-btn-fullbleed'];
+let button;
+
+// Loop out all widths
+for (let j = 0; j < fullbleed.length; j++) {
+  // Loop out all sizees
+  for (let i = 0; i < size.length; i++) {
+    console.log(
```
gaonita, line 10: these 'button playground' files are just for the test with this PR, I assume? 
but having a page(story) like this can be nice indeed like carbon's case you shared

gaonita (commented):
- `npm run chromatic` works but it' skpping most of stuff, like a screenshot below. wonder it is how it is set up for now.

image

- I could access build via the [url]( https://www.chromatic.com/manage?appId=620cc1abd4e7e7003aa19d3e&view=collaborate) 

left few comment here and there, AWESOME initiative @areblk !

areblk:
> I now understand what the problem with paddings and left alignment was. However, it still seems to be that storybook component and reference design are not placed at the same place, so.. is this good enough for us?

I'm on a separate task to place component top left in the storybook, except for stories that for some reason needs to have a different layout. There will be some pixels difference, I guess, but after the change it will be good enough.

areblk:
> * `npm run chromatic` works but it' skpping most of stuff, like a screenshot below. wonder it is how it is set up for now.
> 
> image
> 
> * I could access build via the [url](https://www.chromatic.com/manage?appId=620cc1abd4e7e7003aa19d3e&view=collaborate)
> 
> left few comment here and there, AWESOME initiative @areblk !

Chromatic seems to be running a "pre-test" to see if there are changes before it starts to make new snapshots. If there are no changes, there won't be any new snapshots.

package.json
areblk: I think I added that to the wrong file. It has been removed in other package.json because it makes the whole thing freeze.

demo/HTML/button-playground.js
areblk: it's actually for the HTML demo, to create a complete set of buttons, but it didn't work. I'll remove the file. I do think it's interesting to create stories of button groups, maybe we can do that in the future.

gaonita (commented):
It seems that all the changes on package-lock.json files under theme package are not relevant for this PR, right? if that's the case, let's remove them from the PR.

areblk:
> It seems that all the changes on package-lock.json files under theme package are not relevant for this PR, right? if that's the case, let's remove them from the PR.

I think they're not relevant. How do I remove from commit/PR without actually removing the file?

areblk:
Too many irrelevant and unrelated changes in files. New branch with new PR coming up.

PR #311, opened 2022-02-18
refactor(storybook):increased testability

areblk:
**Describe pull-request**  
- Set 'padded' as global default layout for stories so it's easier to manually check visual appearance
- Set 'centered' as layout for components needing this (tooltip, spinner)
- Remove excess demo code, such as margins and padding, to avoid false positive or negative visual test results
- Slight refactor on tooltip stories to increase testability
- Adding more Lorem ipsum to accordion to better visualise what the paddingReset-control does
- Components that doesn't have a default width has been set to 400 for demo purposes (card, slider, textarea, accordion)
- Change control types from select to radio to make control values more clear

Bonus:
- Moving fullpage-pattern story from /theme to /patterns so it's easier to find. (When refactoring header and side menu, we've forgot to update these stories which made them break)

**Solving issue**  
Fixes: [AB#1678](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1678) [AB#1693](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1693)

**How to test**  
1. Go to storybook preview
2. Look through all components
3. Notice if layout should be different on any of the components
4. Notice if control types are either text, toggle or radio

**Screenshots**  

image


components/src/components/toggle/toggle.stories.js
```diff
@@ -43,28 +43,18 @@ const Template = ({ size, disabled = '', headline = '' }) => {
       ? `<div class="sdds-toggle-headline">${headline}</div>`
       : '';
   return `
-    <style>
-    .demo-container{
-      width: 500px;
-    }
-    .demo-toggle {
-      margin: 5rem 2rem;
```
gaonita, line 51: need some margin between two toggles 
Screenshot 2022-02-18 at 17 20 19

components/src/components/accordion/accordion.stories.js
```diff
@@ -44,19 +44,20 @@ export default {
 
 const Template = ({ disabled, affix, paddingReset }) => `
   <sdds-theme></sdds-theme>
-  <div class="container-demo" style="width:500px; padding:var(--sdds-spacing-layout-48);">
```
gaonita, line 47: _Components that doesn't have a default width has been set to 400 for demo purposes (card, slider, textarea, accordion)_ -> the width for accordion is not set

Screenshot 2022-02-18 at 17 21 33

areblk: This is fixed in latest commit. I had to refresh preview link to be able to see it in amplify.

components/src/components/toggle/toggle.stories.js
areblk: Agree, but it should be set somewhere else than in the demo so it reflects for other users. Can you or @theJohnnyMe help me? :)

gaonita: how do you mean? why not in the demo?

areblk: If it's only in the demo, and not in the component styling, this will give a 'false positive' result on whether the component is correctly styled out of the box. Although, there is no design rule for margin or padding when they're stacked.

gaonita: ah, you meant we need to include margin in the component? It's not necessary because users should be able to set the margin between toggles or with other components on their demand. 
anyway how about having only one toggle on the story? @theJohnnyMe suggested and I think that makes more sense!

gaonita (approved):
Looks good to me 👍🏻

areblk:
> Just a reminder on the continuousSlider one 🔝 , after that one is fixed it can be merged.

Not sure what part of continuous slider that should be tweaked to get the result we want. Could this be done in a different task/PR since it impacts the actual component and not the presentation of it in Storybook?

theJohnnyMe:
> > Just a reminder on the continuousSlider one 🔝 , after that one is fixed it can be merged.
> 
> Not sure what part of continuous slider that should be tweaked to get the result we want. Could this be done in a different task/PR since it impacts the actual component and not the presentation of it in Storybook?

Sure, just make a ticket for it so we do not forget about it.

PR #316, opened 2022-03-03
fix(fullpage): temporary fix for fullpage

gaonita:
**Describe pull-request**  
temporary fix for the fullpage. The header is Updated to new one.

**Solving issue**  
Fixes: broken header on storybook fullpage example

**How to test**  
1. Go to storybook pattern s-> fullpage example
2. Check if the example page looks fine compared to the current one

**Additional context**  
This is temporary fix with new header. We will need to update the sidemenu there when the refactoring is done.


theJohnnyMe (approved):
Looks good to me.

Issue #278, opened 2022-01-12
Improvement - Allow for other min-width for sdds-textfield


blomqvist:
**Describe current situation**  
The current `min-width` for `<input>` is `208px` as of now. For tables with many inputs, this will make for a very wide table.

**Describe the improvement**  
Have a default `min-width`, but allow perhaps a few variants to be set via `<sdds-textfield width="sm">`, or just to be able to supply a `rem` based value; `<sdds-textfield width="10">`. (Or perhaps apply the `input[size]` attribute?)

**Screenshots**  
My use case is a table with inputs that only requires 1-3 digits. There is a lot of white space in the inputs making my table wide.
![bild](https://user-images.githubusercontent.com/163365/149120806-8dee3edb-67f5-49b1-99a0-1e302a2254da.png)

[AB#1381](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1381)

PR #318, opened 2022-03-04
Add an option for a smaller min-width on textfield

blemko:
**Describe pull-request**  
Adds an option for a smaller `min-width` on the textfield component. The default is un-touched.

**Solving issue**  
Fixes: #278 (and perhaps also [AB#1381](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1381)).

**How to test**  
Create a textfield, and supply `width='small'` to get a textfield that has a `min-width` of `104px`
`width='default'` or doing nothing should set `min-width` to `208px`.

**Screenshots**  
bild
`width='small'` to the left and the default to the right.

**Additional context**  
I'm not sure on how this project prefers to handle these kinds of customizations, but I guess this can serve as a template at least.

(tested in Storybook; `min-width: 104px` is somewhat respected, but other properties seems to set the `width` to `180.5px`


theJohnnyMe:
I will check this one on Monday, thanks for your contribution, much appreciated! 🍾

alexlioliopoulos:
Let me talk about this with the designers, before merging anything.

areblk:
I think this is an interesting feature since Scania most likely use short data in many applications, like vehicle registration plates, mileage, salary, time, dates etc. Although, this addition have some issues:

1. The min-width attribute doesn't set the width of input field to 104px. Width attribute value needs to be set somewhere else in the component. Can't really figure out where without changing attribute to fixed width of 104px.
2. Size names apply to component height in this case. I'd suggest that a less wide variant of textfield could be named 'short' instead of small. Or something else that fits naming convention of classes.

theJohnnyMe:
We talked about this within our team. We agreed to solve it by setting the width to 100% and leaving the min-width to 204px. We would add a prop called noMinWidth (boolean) so it would disable the min-width rule (if set to true) and the element will be wide as much as its parent container allows it. The component is already integrated into some places so we do not want to mess up current implementations, but still enable both current and new users with new feature/option of component. I hope so you will find it useful @blemko .

 I think that soon we could fix it, or if you feel like and you have time you can submit new PR that will fix it in upper mentioned way as well.

blemko:
So.. at last. Perhaps a bit messy commit history - I guess that can be squashed.

A new property is implemented, `minwidth`, which defaults to `true`. I'm not so comfortable with Storybook and Stencil, so I could not manage to get a non lowercase property to work (I do realize, as I'm writing this, that perhaps the property gets translated to kebab case?).

blemko:
> Hi buddy, one of our colleagues picked up an issue you have been working on, I think it might go out on Monday. Please check here: #348

ah, alright! I guess we can close this one then :)

theJohnnyMe:
> > Hi buddy, one of our colleagues picked up an issue you have been working on, I think it might go out on Monday. Please check here: #348
> 
> ah, alright! I guess we can close this one then :)

Please stay tuned for our release on Monday. I hope the solution will fix the issue you faced. :) 
Thanks again for your goodwill to contribute and improve our code.

PR #320, opened 2022-03-04
fix(storybook): set global sdds-theme instead of in every story

theJohnnyMe:
**Describe pull-request**  
Moving <sdds-theme> to the global import in Storybook

**Solving issue**  
No ticket for this one I think.
This is dev-ex improvement: Story file no longer needs <sdds-theme> to be imported just above component code. Now it is globally consumed - pretty much the same way as our users do when they set up their code to consume our design system..

PS: Added tow commands in package.json of components to test lighter start up.

**How to test**  
1. Just check if all stories render as usual.

### DIFF
--- a/components/.storybook/preview-body.html
+++ b/components/.storybook/preview-body.html
@@ -4,3 +4,6 @@
         max-width: 460px;
     }
 </style>
+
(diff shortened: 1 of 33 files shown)


areblk (approved):
Storybook stories render as expected.

PR #325, opened 2022-03-11
fix(dropdown): post integration improvements

theJohnnyMe:
**Describe pull-request**  
Fixes made during integration test

**How to test**  
1. Go to Storybook
2. Check that the dropdown component looks and behaves as before

### DIFF
--- a/components/src/components/dropdown/_dropdown-core.scss
+++ b/components/src/components/dropdown/_dropdown-core.scss
@@ -35,7 +35,7 @@ html {
     &::before {
       content: ' ';
       position: absolute;
-      z-index: 3;
+      z-index: 2;
       bottom: 0;
       left: 0;
       width: 100%;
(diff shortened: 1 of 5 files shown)


areblk (approved):
PR looks better because now icon stays the same size also for when the selected option is truncated.

areblk:
Another note. I'm not sure about the centered layout for the dropdown in storybook. Can you think of reasons to keep it centered? @theJohnnyMe

theJohnnyMe:
> 

I did it so I can easily test upwards/downwards menu opening. It was easier to select the option if there was some space left above the dropdown. In padded layout, a menu opens upwards but it felt trick to select any option. If it messes up testing in chromatic let's consider adding a wrapper around it with some 50-100px margin/padding to the top that will always be there so the upwards menu open can be tested in an easy way.

areblk:
> 

Ah, I see what you mean. It's not in chromatic yet, so it doesn't matter for that sake. What I was thinking about is that right now, there's no good way in storybook to control for how the options are being presented since it's always in the center. Let's leave it like this now and I'll think of ways to tell the story to force it opening upwards or downwards :-)

PR #349, opened 2022-04-06
feat: accordion-item toggle event emitter

bogdan-todireanu:
Added an event named "accordionItemToggle" on the accordion item component which emits a boolean each time after that accordion item is opened/closed.


theJohnnyMe (approved):
Thanks for the contribution Bogan, it looks good to me.

PR #356, opened 2022-04-13
fix(modal): there is a issue setting overflow:hidden on body 

mmexvr:
**Describe pull-request**  

Opening the modal in a page with a height over the screen size makes for example the header height
smaller. What I have done is switch the overflow:hidden so it is on the html instead of body element.

This problem exist even without modal, if you set the body with overflow hidden

**How to test**  
1. Open a page with a modal, the header component and a page that have the height bigger then screen (scrollbar)
2. Add the modal
3. Open them modal and see the header(video below)

**Screenshots**  

Problem

https://user-images.githubusercontent.com/35451568/163327938-cbb26468-cb39-47b0-9ea8-c5b53d98a792.mov

Working example after

https://user-images.githubusercontent.com/35451568/163328061-1cabba54-cf0e-4aab-9bda-86744eb77bee.mov

### DIFF
--- a/components/src/components/modal/modal.tsx
+++ b/components/src/components/modal/modal.tsx
@@ -83,7 +83,6 @@ export class Modal {
   render() {
     return (
       <Host class={`sdds-modal-backdrop ${this.show ? 'show' : 'hide'}`}>
-        fdsafdsafdsafdsafdsa
         <div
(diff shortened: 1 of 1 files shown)


areblk (commented):
Not noticing any regressions in Storybook from this change

theJohnnyMe (approved):
Looks good to me ;)

gaonita (approved):
LGTM 👍🏻

PR #358, opened 2022-04-14
Task 1874

victorscania:
New webcomponent: datetime

this webcomponent uses the html-native datetime, date and time-picker
Fixes: [AB#1874](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1874)

**How to test**  
Open storybook and select the datetime component

**Additional context**  
This not a final PR, but intended to demo the component in its current state


areblk:
Tested on Edge and Windows

# Issues
Although most testing is done on datetime-local type, issues are also valid for type=date and type=time
1. **Basic.** (Only valid for Type=datetime-local). Input field is too short to show selected date and time. When clicked again, "PM" disappears and shows a somewhat correct value but without PM or AM.
image
2. **Disabled.** Icon should have the same color as text to make it more apparent that it's disabled https://digitaldesign.scania.com/components/datepicker/usage#states
image
3. **Labels.** Cannot see where the label goes. By the inspection tool, Storybook seem to pass it to the DOM but I cannot see it.
image
4. **Labels > Label inside.** Label is displaced when no date is selected. (Label looks OK when date is selected)
image
5. **Helper.** Helper text works in default state, but disappears when ex. label is added inside. Shadow DOM looks a bit corrupt
image
image
6. **State > Error state.** Should have an icon before Helper text. (See how it's done in the Dropdown component https://deploy-preview-358--sdds-storybook.netlify.app/?path=/story/component-dropdown--helper&args=state:error)
7. **State > Error state.** Calendar/clock icon should be red when in error state https://digitaldesign.scania.com/components/datepicker/usage#states.
8. **Icon.** Not able to see where icon goes.
image

# Note
9. Format according to SDDS should be YYYY-MM-DD, and preferably also 24H format on time, but I find no easy access evidence that it's possible using the native one.

components/src/components/datetime/datetime.scss
```diff
@@ -0,0 +1,374 @@
...
+  }
+}
+
+// icons
+
+.sdds-datetime-input,
+.sdds-datetime-input-md {
+  &[type="datetime-local"] {
+    ~ .icon-datetime-local {
+      //background-color: red;
```
theJohnnyMe, line 46: Maybe can be removed as no rule is assigned?

components/src/components/datetime/datetime.scss
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    ~ .icon-time {
+      display: none;
+    }
+  }
+
+  &[type="date"] {
+    ~ .icon-datetime-local {
+      //background-color: green;
```
theJohnnyMe, line 56: Same here.

components/src/components/datetime/datetime.scss
```diff
@@ -0,0 +1,374 @@
...
+    }
+  }
+
+  &[type="time"] {
+    ~ .icon-datetime-local {
+      display: none;
+    }
+
+    ~ .icon-time {
+      //background-color: blue;
```
theJohnnyMe, line 70: Same here.

components/src/components/datetime/datetime.tsx
```diff
@@ -0,0 +1,198 @@
...
+  /** Change event for the datetime */
+  @Event({
+    composed: true,
+    bubbles: true,
+    cancelable: true,
+  })
+  customChange: EventEmitter;
+
+  // Listener if input enters focus state
+  @Listen('focus')
```
theJohnnyMe, line 65: Nice!

components/src/components/datetime/datetime.tsx
```diff
@@ -0,0 +1,198 @@
...
+  handleFocusIn() {
+    this.focusInput = true;
+  }
+
+  // Listener if input leaves focus state
+  @Listen('focusout')
```
theJohnnyMe, line 71: Nice!

theJohnnyMe (changes requested):

theJohnnyMe:
overlapping

Placeholder and label overlapping. I think Annika mentions similar/same issue under no.4.

theJohnnyMe:
1. See if it is possible to add a prop for min & max date so the user can specify that too.

2. Another useful thing is to present the selected date&time as seconds in the attribute for internal calculation.
It could look like:
`<sdds-datepicker value="2017-06-01" numeric-value="1496275200000" ... > `

alexlioliopoulos:
@victorscania What are your feelings about the issues identified by Annika and Johnny? I mean, is there anyone that would require long time to be fixed, hence we should consider if it should be prioritized or not? Let me know!
But anyway, as always, prio #1 are things that look broken. In this case the "label-inside" variant. Styling problems or extra functionality come after that.

areblk:
# Fixed
- (2) Icon is grey when disabled

# Status
Everything else as before, except for
- Icon prop, because that has been removed
- Label inside, has also been removed

My opinion on what's to be considered broken and therefore should be fixed are things having to do with labels and error helper text (3, 4, 6). I'd prioritize nr.3 (making label text visible) because all of my other points, including 4 and 6, can be solved with workarounds in consumer apps.

alexlioliopoulos:
> 

I would say let's prioritize 3 & 4. If our users are using the label-inside textfield in a form, they need to have the equivalent for the dropdown. 

Another thing to work on, is to create a "small" variant that has a height of 40px (large is 56px tall, medium is 48px tall).

areblk (approved):

PR #368, opened 2022-05-11
feat(table): multiselect

theJohnnyMe:
**Describe pull-request**  
Creating Multiselect feature for data table component

**Solving issue**  
Fixes: AB#1697

**How to test**
1. Go to Storybook
2. Check Data Table component -> Web Component
3. Play around multi-select feature (select one, deselect, select all, deselect one)


alexlioliopoulos (approved):
From my side, looks good! As i told you before, I had some concerns about the column hover on top of a selection, but I don't think we can find a really good solution to it, so I am happy with how it is right now.

areblk:
# Just notes, no request for change
## Checkboxes
- Table header checkbox could be considered a ['parent checkbox' which in docs is described](https://digitaldesign.scania.com/components/checkboxes/usage#types) like this: *When the parent checkbox is selected, all child checkboxes are selected. If it unselected, all child checkboxes are unselected. And if some of the child checkboxes are selected, the parent checkbox becomes an indeterminate checkbox.* Currently, sdds doesn't feature a indeterminate variation of checkbox. 
- Not possible to operate checkboxes using a keyboard only. 

## Data table as web component
- No apparent issues with z-index when placing a sdds-dropdown underneath the table. I noticed that it's not possible to place another component within the data table, which is think is something we should look into. This would allow users to have input elements within the web component data table, when they like to edit data presented there. Use case: click edit on a row or cell to edit it's content and save to database/json.

areblk (changes requested):
## Data table multiselect
1. When space is narrow, checkbox in `.sdds-table__header-cell sdds-table__header-cell--checkbox` is very tight. I suggest to make it minimum with 24 px to fit checkbox, hover/focus outline and match it with available spacing values

image

** EDIT: ** minimum width of checkbox column should be same as its height. Default: 48px, compact: 32px excl. border. Checked with Alex and measured from design reqs.

areblk (approved):

PR #369, opened 2022-05-11
fix(textfield): height improvements

samuellindsoleil:
Hello!

In this PR i've removed the fixed height on textifelds. With this improvement we can easliy reuse the component in other projects where we need to adjust the height. 

Please let me know if i missed something.

### DIFF
--- a/components/package-lock.json
+++ b/components/package-lock.json
@@ -1,8 +1,28135 @@
 {
 	"name": "@scania/components",
 	"version": "4.1.0",
-	"lockfileVersion": 1,
+	"lockfileVersion": 2,
 	"requires": true,
+	"packages": {
+		"": {
+			"name": "@scania/components",
+			"version": "4.1.0",
+			"license": "MIT",
+			"dependencies": {
+				"concurrently": "^5.3.0"
+			},
+			"devDependencies": {
+				"@babel/core": "^7.17.8",
+				"@babel/plugin-syntax-jsx": "^7.16.7",
+				"@babel/plugin-transform-react-jsx": "^7.17.3",
+				"@popperjs/core": "^2.11.4",
+				"@stencil/core": "2.14.2",
+				"@stencil/router": "1.0.1",
+				"@stencil/sass": "^1.5.2",
+				"@stencil/store": "^1.5.0",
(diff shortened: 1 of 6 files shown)


alexlioliopoulos:
Hi, thanks a lot for taking the time to do this! We will have a closer look at this one soon (I am in the office and Storybook is not accessible from the scania network). 
A couple of things before having a look though:
In a couple of days we are going to release a Small variant for our textfield component, that hopefully will make it possible to use it in more compact spaces.
The reason why we don't leave full freedom on the height of this component, is to avoid stretching its use, just for the sake of fitting things in small spaces, disregarding usability or UX considerations. Also, by only having 3 size variants, we can achieve higher consistency across different apps/pages within an app.
Would you say that you need considerably smaller textfields? Our large one is 56px high, our medium one 48px high and the new small variant will be 40px tall.

samuellindsoleil:
Thanks for replying!

Sounds good that you will be making another size for the textfield. I understand your thoughts and ideas why you want fixed heights. However though, it will only work when you're using your library from scatch in a new project. If you're migrating from an already existing project you need to make bigger adjustments and it will affect the application on greater matter. I'm looking for a textfield with a height of 30px-35px. We could indeed use your textfield as it is today but in our case we need to rebuild the hole header just to make it work. What do you think about a prop for the height? We could use 50px as default and perhaps a minimum of 30px? @alexlioliopoulos

alexlioliopoulos:
We are still a bit hesitant about opening up the component in this way, for the reasons I explained in my previous comment. If including such a prop, hence making it super easy for consumers of our packages to go super-small in the input elements, we expect many of them to go that way for most of the cases (not only specific cases like the header one that you are mentioning, which makes sense). Using smaller things seems to be appealing, since you can gather more things in a small real estate. However, it usually brings worse UX. 
So, regarding your specific issue, I would suggest creating a new component that would resemble the styling of the SDDS textfield one, and make it as tall as you need it to be. Or of course, use the SDDS header as well, which could work fine with the textfield too. What do you think?

samuellindsoleil:
Yes you've got a point. We will build our own component instead as you suggested. We can use your textfield-component in the future when we're ready for your header as well. We can close this PR. @alexlioliopoulos

PR #372, opened 2022-05-16
chore: release 31

theJohnnyMe:
**Describe pull-request**  
Release PR for sprint 31


areblk (approved):

areblk:
My only comment is that we could remove story of patterns-fullpage because 1) it is broken, 2) that story is contained within side menu story. Everything else as expected!

theJohnnyMe:
> 

I will remove it in a separate PR. Good point ;)

PR #373, opened 2022-05-17
feat(data-table): test search capability

theJohnnyMe:
**Describe pull-request**  
The search feature of the data table

**Solving issue**  
Fixes: AB#1858

**How to test**  
1. Open Storybook
2. Go to Table --> Web component
3. Try to type in some string in input to see if the table will filter rows that contain it.

**Additional context**  
Under development, still many things to sort of in the graphic part, code and story file.

### DIFF
--- a/components/src/components/table/table.tsx
+++ b/components/src/components/table/table.tsx
@@ -81,9 +81,13 @@ export class Table {
 
   @State() bodyDataManipulated = [];
 
+  @State() bodyDataOriginal = [];
+
   @State() bodyRowSelected = false;
 
   @State() multiselectArray = [];
+
+  @State() arrayOfKeys = [];
 
   @State() multiselectArrayJSON: string;
 
@@ -103,6 +107,8 @@ export class Table {
       this.innerBodyData = newValue;
     }
     this.bodyDataManipulated = [...this.innerBodyData];
+    this.bodyDataOriginal = [...this.innerBodyData];
+    this.arrayOfKeys = Object.keys(this.bodyDataManipulated[0]);
   }
 
   // Listen to sortColumnData from table-header-element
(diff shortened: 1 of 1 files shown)


alexlioliopoulos (commented):
First of all, it wild to see the component being on steroids!!!!

A minor thing, when I press Sonya, the alignment kinda gets messedup, however I feel like this is a storybook problem and not a component problem.

image

theJohnnyMe:
> **Looks great overall!** @theJohnnyMe
> 
> _It would be good if we fix the following before releasing:_
> 
> 1. When making a search and then sorting, things get messed up a little bit. Could we disable sorting functionality after a search has been performed? If that is easier, before we do a more stable fix for this.
> 
> _Optional things that could be worked on in the future:_
> 
> 1. Changing the pointer icon from the "line" that is now, to a "hand", when the search bar is collapsed.

Thanks for the feedback Alex. Let me then try to make it like this:
If there is a search string - disable sorting, otherwise enable it.

Icon to hand should be a quick one. Let me try to sort it out.

theJohnnyMe:
@alexlioliopoulos - I think I got it sorted, please check after build is done ;)

alexlioliopoulos:
image

alexlioliopoulos (approved):

components/src/components/table/table.tsx
```diff
@@ -239,6 +292,79 @@ export class Table {
       </tr>
     ));
 
+  // ToDo: below is just a test concept
```
gaonita, line 295: Does 'test concept' mean that we are not using these functions now but planning to enable them in near future?

gaonita (approved):
great work 🙌🏻
and awesome that you added lots of possible logics for the future!

components/src/components/table/table.tsx
theJohnnyMe: It is just a comment that I forgot to remove, now a concept becomes a reality.
I will drop that one as it won't be needed anymore.
Thanks for the review and approval!

PR #377, opened 2022-05-25
fix: package files update

theJohnnyMe:
**Describe pull-request**  
Adding files that I missed to build during release so it creates issues for Amplify work.


gaonita (approved):
👍🏻👍🏻👍🏻

PR #378, opened 2022-05-31
fix: amplify and math div issue

theJohnnyMe:
**Describe pull-request**  
Fixing error that comes from SCSS files where we use "/" to divide. In the migration manual, it is proposed to use math.div or wrap math expressions in calc() .
Another issue to fix was to remove all package-lock files + node_modules to run the clean install as Amplify was failing to build properly since release 29.


gaonita (approved):
Thank you for doing such chores 👍🏻  it is getting cleaner and cleaner 🧼

PR #381, opened 2022-06-02
feat: chips & message components

victorscania:
**Describe pull-request**  
Two new components. Chips and Messages

**Solving issue**  
Fixes: [AB#1730](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1730) and [AB#1904](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1904)

**How to test**  
1. Go to the storybook
2. check chips, go thru the storie and see if they are correct
3. check messages, check the stories, and use the controls to se if you can create all variants.


albinscania (approved):
Looks good and works great.
It's weird that your commits however seem to remove quotes and add double quotes instead... :)

components/src/components/message/message.scss
```diff
@@ -0,0 +1,172 @@
...
+        }
+      }
+    }
+
+    &-error {
+      color: var(--sdds-grey-958);
+      background-color: var(--sdds-red-50);
+
+      &::before {
+        background-color: #ff2340;
```
gaonita, line 124: these semantic colors as well, we can use semantic variables? 🚥

components/src/components/message/message.scss
```diff
@@ -0,0 +1,172 @@
...
+    background-color: #2b70d3;
+  }
+
+  &-icon {
+    height: 20px;
+    width: 20px;
+    position: absolute;
+
+    //top: 16px;
+    //left: 20px;
```
gaonita, line 46: found commented out lines 😬

components/src/components/message/message.scss
```diff
@@ -0,0 +1,172 @@
...
+
+  &::before {
+    content: " ";
+    display: block;
+    position: absolute;
+    width: 4px;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    background-color: #2b70d3;
```
gaonita, line 37: can this color be a variable like '--sdds-blue-400' maybe?

gaonita (approved):
both work great! I just found some tiny stuff, left some comments here :)

and yeah like Albin says it looks like Victor's prettier(?) and Johnny's eslint(?) fighting against with single/double quotation mark 😆

theJohnnyMe:
> Looks good and works great. It's weird that your commits however seem to remove quotes and add double quotes instead... :)

I will check the settings on my machine. Something is off for sure :D

components/src/components/message/message.stories.js
```diff
@@ -0,0 +1,151 @@
...
+      break;
+  }
+
+  const extendedMessageHtml = `
+    <p class="sdds-message-extended">
+      Longer message text can be placed here. Longer message text can be placed here. Longer message text can be placed here.
+    </p>
+  `;
+
+  const iconHtml = `
```
theJohnnyMe, line 103: I suggest turning this SVG to a button holding SVG that has a "pointer" cursor when hovered over:

Screenshot 2022-06-03 at 09 44 45

That way we will deliver a ready solution so users can easily attach a function to that button.

theJohnnyMe (changes requested):

components/src/components/chips/chips.stories.js
```diff
@@ -0,0 +1,93 @@
...
+
+    case 'Icon right':
+      iconClass = 'sdds-chip__icon-right';
+      break;
+
+    default:
+      break;
+  }
+
+  const iconSvg = `
```
theJohnnyMe, line 60: Maybe this could be also a button with SVG in it, same as in the message component to easily attach function on it?

theJohnnyMe (changes requested):

components/src/components/message/message.stories.js
victorscania: I dont think that is suppose to be a button rather just an icon. I just used the X icon as an example. Maybe replace it with something else tho, to avoid confusion.

components/src/components/chips/chips.stories.js
victorscania: same thing here. its just an icon. I really should have used another icon ;)

components/src/components/message/message.stories.js
theJohnnyMe: Aha, ok, I got it now. Yeah, "x" confused me. :D 
I checked with @charlottelorensson and each type of message has a dedicated icon. For example, success/positive one has a checkmark there. Can you please check with her which icons are associated with which type and fix it in the example in Storybook?

components/src/components/chips/chips.stories.js
theJohnnyMe: All good here then ;) I agree with you, maybe, just as a suggestion try to add an example with "Like" as text and a "Hearth" icon next to it to avoid confussion. I feel others will end up in similar trap like I did hehe

theJohnnyMe (commented):
@victorscania Looks good to me, thanks for sorting icons confusion ;)
One thing is to check Gaon's comments, maybe you missed them, I think it is a quick fix.

alexlioliopoulos (changes requested):
They both look great!

I only have one request: 
We should add a small variant for the chips. The large one is 32px high and the small one is 24px high. Otherwise they are identical.

gaonita (approved):
LGTM 👍🏻

alexlioliopoulos (approved):

PR #382, opened 2022-06-02
feat(table): actionbar section

theJohnnyMe:
**Describe pull-request**  
Add a slot in the header of the table so a user can add additional button/s for their own functions. 

**Solving issue**  
Fixes: [AB#1858](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1858)

**How to test**  
1. Open Storybook
2. Select the WebComponent version of it
3. Enable "Actionbar" from Storybook controls to see it appearing

**Additional notes**
In order to add any kind of HTML element to action bar user has to only add attribute slot="sdds-table__actionbar" to it. Check story file in your code editor or here in GitHub to understand what is happening. 
On sdds-website, it would be good to advise users to only buttons as they trigger some king of action. 

**Question to think about**
Right now we developed a special sort of button styling just for a table. The question is do we really want to do this or promote usage of our own buttons in small size there, for example, small button in the ghost variant is similar to the newly designed button in [Figma proposal](https://www.figma.com/file/ZUT3Te7JVWggoukL9QGsPp/Components?node-id=4%3A6).


albinscania (approved):
Looks good!
And you're right about the clone button. Might be something we can look at in the future, but for now I think it works fine!

PR #383, opened 2022-06-07
feat(table): pagination

theJohnnyMe:
**Describe pull-request**  
Pagination feature of data table web component

**Solving issue**  
Fixes: [AB#1941](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1941)

**How to test**  
1. Go to Storybook --> Table --> Web Component
2. Test pagination feature


components/src/components/table/table.scss
```diff
@@ -212,6 +212,114 @@
...
+
+        .sdds-table__page-selector-input--shake {
+          animation: sdds-shake 0.5s cubic-bezier(0.36, 0.07, 0.19, 0.97) both;
+          transform: translate3d(0, 0, 0);
+          backface-visibility: hidden;
+          perspective: 1000px;
+        }
+
+        // ToDo: Fina a better place for animation ? Johnny
+        @keyframes sdds-shake {
```
gaonita, line 261: 🤔 just a question: Isn't this a good place to have animation? or do you have better idea in your mind?

gaonita (approved):
pagination works like a charm 🙌🏻

components/src/components/table/table.scss
theJohnnyMe: It is good now, but for example, maybe in the future another component would like to have the same one, then maybe it would be a good idea to move it over to some general animation file, let's say __sdds-animation.scss so it is accessible from other places. Similar to typography and colors today. Thanks for the review! ;)

gaonita: aha I understand. yes of course! awesome that you are always foreseeing the future 🧙🏻‍♂️

PR #384, opened 2022-06-09
feat: readonly prop for textfield & textarea 

victorscania:
**Describe pull-request**  
Added read only settin to textfield and textarea

**Solving issue**  
Fixes: [AB#1947](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1947)

**How to test**  
Check the textfield and textarea on the storybook-link


theJohnnyMe:
@charlottelorensson and @alexlioliopoulos I think we have to take long text as a possibility here, please observe these two edge cases, one for each:

Screenshot 2022-06-09 at 15 50 10

Screenshot 2022-06-09 at 15 52 44

My two cents are: Move the read-only icon closer to the top-right edge of the screen. Possibly make it a bit smaller and maybe use a lighter shade for its color, for example, grey-600 or 700 ? What do you think?

alexlioliopoulos (changes requested):
Read-only text should **not** have a background. If I am not mistaken, here we keep the background and we only add an non-edit icon. Have a look at the task in Azure and the Figma prototype to check what I mean.

https://dev.azure.com/scaniadigitaldesignsystem/SDDS/_sprints/taskboard/Product%20Team/SDDS/Sprint%2034%20-%20Hrvatska?workitem=1947

Also, the ideas was to have a tooltip on hover of the non-edit icon. Is that not possible to have?

alexlioliopoulos:
> 

Good point! I think that we shouldn't change the placement of the non-edit icon, since it now aligns well with the text. Moving it up might create a sense of unbalance. But we should certainly fix the issue you are bringing up. 

Would it be possible to block content from going behind the icon area? This is how Carbon is doing it:

image

So pretty much the icon area has a bg behind it and nothing can go behind this. But I guess that might be a bit complicated, or?

alexlioliopoulos (changes requested):
Great changes! I have a couple of more comments:

- The field can actually be editted, if users click on the icon.
image

- In the label-inside variant, if we double click on the field the placeholder text kinda appears:
image

victorscania:
> Great changes! I have a couple of more comments:
> 
> * The field can actually be editted, if users click on the icon.
>   image
> * In the label-inside variant, if we double click on the field the placeholder text kinda appears:
>   image

Super weird interaction. Seems to work as intended on the textarea, but not on the textfield even tho its the same solution on both. The icon seems to kinda work like a backdoor to access the textfield. I have no quick fix solution for this now, gonna have to poke it a bit to figure it out.

alexlioliopoulos:
Idea: Could we just not use the icon, as a quick fix until we find a better solution?

victorscania:
> Idea: Could we just not use the icon, as a quick fix until we find a better solution?

Or maybe remove the popup "this field is non-editable" label and just use the icon? thats what causing the issue in the first place.

alexlioliopoulos:
> > Idea: Could we just not use the icon, as a quick fix until we find a better solution?
> 
> Or maybe remove the popup "this field is non-editable" label and just use the icon? thats what causing the issue in the first place.

Works for me!

theJohnnyMe:
@victorscania and @alexlioliopoulos  - What is the status on this one?

alexlioliopoulos:
I guess we are waiting for Victor to make the changes, right?

PR #386, opened 2022-06-14
fix(tabs): remove underline on hover state when showing tabs

victorscania:
**Describe pull-request**  
fix: (tabs) fixed underline on hover state when showing tabs

**Solving issue**  
Fixes: [AB#1957](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1957)

**How to test**  
go to the storybook and see if the hover state on inline tabs and navigation tabs have "text-decoration: none;"


theJohnnyMe (approved):
LGTM! As you say, let's test it once released on sdds-website, hard to test otherwise.

PR #388, opened 2022-06-14
fix: renaming and cleaning demo folder

theJohnnyMe:
**Describe pull-request**  

1. Renaming demo to "sandbox" (+ updating all internal paths... )
2. Adding readme file with an explanation of what is a sandbox and why we use it + link to the proper demo page
3. Leaving only Angular version 13 for testing and removing others
4. Adding http-server as a dependency for HTML example

**Solving issue**  
Fixes: AB#1955

**How to test**  
1. Take branch on your local
5. Try to run each environment (HTML/Angular/React/Vue)
6. Check if the sandbox readme file looks OK

Note: @alexlioliopoulos - I do not want to bother with testing the dev environment, just see if sandbox -> readme.md sounds OK to you.


alexlioliopoulos:
Not sure where I should check since the deploy preview sends me in Storybook, but Sandbox sounds good I trust you for the readme.

alexlioliopoulos (approved):

sandbox/HTML/package.json
```diff
@@ -1,18 +1,20 @@
...
   "name": "sdds-html-demo",
   "version": "1.0.0",
-  "description": "Install `http-start` globally on your computer.",
+  "description": "HTML testing environment",
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "link": "npm link @scania/components && npm link @scania/theme-light && npm link @scania/icons"
+    "link": "npm link @scania/components && npm link @scania/theme-light && npm link @scania/icons",
+    "start": "http-server -p 4242"
```
gaonita, line 9: clever!  👏🏻

gaonita (commented):
Tested angular, react, vue and HTML in sandbox, all work fine! 
and each readme makes sense to me 👍🏻

under demo, angular13 src files are gone though 😮 is it intended because the plan is to remove it from the package?

sandbox/HTML/package.json
theJohnnyMe: Thanks for the feedback!
Angular13 one is remaining under the sandbox folder now, old demo folder is removed so all good there.

PR #390, opened 2022-06-14
fix(storybook): removing fullpage example + header,footer,side-menu under component category

theJohnnyMe:
**Describe pull-request**  
Removing fullpage story example as it is not maintained and not needed anymore

**How to test**  
1. Go to the Storybook link once it is built
2. Check if under the "Patterns" category you can see "Fullpage" (it should be removed)
3. Header, Footer and Side-menu are now under Components with other components
4. If all good, great, hit Approve :D


alexlioliopoulos (approved):

victorscania (commented):
Theres seems like there's no pattern category at all in the storybook now.

victorscania (approved):

Issue #395, opened 2022-06-23
DataTable Filtering


Tommy-Hallgren:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have made sure that the issue comes from SDDS packages by removing SDDS from my project as a test.
- [X] I have followed the installation guide.

### Package

@scania/components

### Package versions

4.5.0

### Browser

Chrome

### Framework

Angular

### Version

Angular 13

### Reproduction steps

When using the data-table from SDDS and when filtering option is applied the original resultset is not restored properly when the filter criteria is removed. This is re-produceable on the demo page as well: https://digitaldesign.scania.com/components/data-tables/code

### Code example

<div>....

### Screenshots

Add/Copy screenshot here

### Expected behaviour

original result is shown in datatable after filtercriteria is removed

### Console errors

Add warning/error message

PR #398, opened 2022-06-27
fix(table): reset search results if search query is empty

theJohnnyMe:
**Describe pull-request**  
Table was not resetting search results after removing search criteria 

**Solving issue**  
Fixes: #395 

**How to test**  
1. Open Storybook on the link generated below
2. Go to the data table component and try to search
3. After you remove the search criteria, the table needs to show the original results 
4. Compare it with behavior on [prod Storybook ](https://production.d1g8v8mrkx992n.amplifyapp.com/?path=/story/foundation-colour--brand)where the bug still exists.

### DIFF
--- a/components/src/components/table/table.tsx
+++ b/components/src/components/table/table.tsx
@@ -373,6 +373,7 @@ export class Table {
       if (this.pagination) {
         this.tempPaginationDisable = false;
       }
+
       sddsTableSearchBar.classList.remove('sdds-table__searchbar--active');
 
       dataRowsFiltering.forEach((item) => {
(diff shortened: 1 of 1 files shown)


alexlioliopoulos:
Works great! My only comment is that if you type "b", then you get to see 4 rows of content, instead of the original 2 rows that were shown if you searched nothing. But that is unrelated to the current ticket.

albinscania (approved):
Looks good and works!

theJohnnyMe:
> Works great! My only comment is that if you type "b", then you get to see 4 rows of content, instead of the original 2 rows that were shown if you searched nothing. But that is unrelated to the current ticket.

Good that you brought this one up, I just realized that there was another bug that I fixed. The issue was that once you put on "b" you get all 7 results, although only 4 are having b. If you insert one more "b" and then remove it, you get the correct results. Now that box is fixed too. It was only showing when pagination is ON together with filtering.

PR #406, opened 2022-07-11
Dropdown fixes sprint 36

EmelieLitwin:
**Describe pull-request**  
Improvements on the dropdown component.
- Added arrow navigation
- Fixed disable option bug
- Increased z-index to fix overlapping  issues
- Closing dropdown when clicking on open dropdown
- Fixed bugs for inside label (Long labels doesn't align with selected option and overflows)

**Solving issue**  
[AB#1994](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1994) 
[AB#1997](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1997) 
[AB#2017](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2017) 
[AB#2014](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2014)

**How to test**  
1. Open storybook
2. Check if arrow navigation works on dropdown component (not on disabled version)
3. Try closing it by clicking on it
4. Try long labels in the inside label option and see if it works correctly


alexlioliopoulos:
Nice! Super fast as well! 

I have nothing to comment on when it comes to the "collapse when clicking outside", "disabled state" and "z-index issue". Although the latter one needs to be tested in contexts that coexist with other elements to make sure it works.

Regarding the "up and down arrows", it works great, however it now does not have a hover state, rather than goes to the focused state while hovering (meaning that you get the blue outline which is the focused state visual hint while hovering, instead of the plain background colour change).

Some further comments for the dropdown, that don't have to do with the work you have done though, but since you are touching upon it, perhaps you might have ideas how to tackle them:

1. There are multiple styling differences between the component as it is designed, and the one in code. 

image

The padding to the left is 16px in design but 20px in code. Also, as can be seen in the image above, the spacings of the elements within the component (label, text etc) are different in design and in code.

2. Sometimes, part of the menu list is not accessible to the users when inside blocks or divs or or. For instance the image below

image

I cannot select the 3-4 first menu items while in my laptop, unless I change the browser zoom to 60%, which is a hack that few people would do though. Is that something that we could somehow solve?

components/src/components/dropdown/_dropdown-core.scss
```diff
@@ -10,7 +10,7 @@ html {
 }
 
 @mixin dropdown-wrapper {
-  @include type-style('detail-02');
+  @include type-style("detail-02");
```
theJohnnyMe, line 13: We can check your IDE for loading settings for ESLint and Prettier. It should be a single quote. Not a biggy.

components/src/components/dropdown/dropdown.scss
```diff
@@ -38,6 +38,10 @@
     align-items: center;
   }
 
+  .sdds-dropdown-toggle--selected:not(:focus) {
```
theJohnnyMe, line 41: Why is :not(:focus) needed for?

components/src/components/dropdown/dropdown.tsx
```diff
@@ -104,19 +108,60 @@ export class Dropdown {
...
       }
-      this.open = true;
+      this.open ? !this.open : this.open;
     } else {
       this.tabbingLabelReset();
       this.open = false;
     }
   }
 
+  @Listen('keydown')
```
theJohnnyMe, line 123: Really neat solution, nice work!

components/src/components/dropdown/dropdown.tsx
```diff
@@ -178,21 +223,23 @@ export class Dropdown {
...
+              this.selectedValue === 'filter' ? 'is-filter' : ''
+            } ${
+              this.selectedValue !== '' ? 'sdds-dropdown-toggle--selected' : ''
             }`}
             type="button"
             onClick={() => this.handleClick()}
             ref={(node) => (this.node = node)}
           >
-            <div class="sdds-dropdown-label">
+            <span class="sdds-dropdown-label">
```
theJohnnyMe, line 242: I guess switching to span as it brings fewer styles to overwrite?

EmelieLitwin: Yeah and semantically it didn't feel great to have a div in a button :)

components/src/components/dropdown/dropdown.scss
EmelieLitwin: I had a reason when I wrote it but maybe it's not needed anymore. I'm going to investigate.

theJohnnyMe:
@EmelieLitwin couple of things I think need some attention:
1. When in Dropdown with the selected option "lable-inside", I think the color of the selected text has to change, a bit darker -  I think Alex mentioned it too.
2. Check disabled state and label color for "lable-inside" variation.
3. Check the disabled state of filter variation, it seems like it does not work.

theJohnnyMe (approved):

alexlioliopoulos:
Great work! Some last minor feedback from my side (plus I think it would be good to add me or a designer as a reviewer as well, for the visual parts):

image

The no-label and inside variants have the text placed a little bit differently, as can be seen in the image above. When inspecting it seems like it has to do with the wrappers (i might be using a wrong term), but seems like a quick fix.

alexlioliopoulos (changes requested):
Great work @EmelieLitwin ! Some last minor feedback from my side (plus I think it would be good to add me or a designer as a reviewer as well, for the visual parts):

image

The no-label and inside variants have the text placed a little bit differently, as can be seen in the image above. When inspecting it seems like it has to do with the wrappers (i might be using a wrong term), but seems like a quick fix.

alexlioliopoulos (approved):
Lovely!

PR #420, opened 2022-08-22
refactor(data-table): separation in sub-components, expandable rows feature

theJohnnyMe:
### DO NOT MERGE - THIS ONE WAITS FOR VERSION 5

**Describe pull-request**  
The new version of data table, it brings:

- separation in sub-components (table, header, header-cell, body, body-row, extended-body-row, body-cell, footer)
- web component can be used as HTML/CSS only too
- disabled functionalities in favour of user own customised solutions
- lots of events 
- event identifier ( so you table can speak to sub-components without interference )
- new variation: expendable rows

**Solving issue**  
Fixes: [AB#1698](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1698) [AB#2023](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2023)

**How to test**  
Simply try to use it.

**Additional context**  
To ease testing, code is released as beta version, check it here: [5.0.0-beta.5](https://www.npmjs.com/package/@scania/components/v/5.0.0-beta.5)

Version id depended on theme-light: 4.4.0


alexlioliopoulos (changes requested):
No need to start with compliments - it goes without saying (you know how much I love this component!).

Straight to some comments:

1. We shouldn't be having an "expand all" icon on the header part. Not according to design at least, plus other design systems are also not having it.

image

2. You are using a "chevron_down" icon of 20x20px size. The size in design is 16x16px, which is also the size we are using for the checkboxes in the other data table variant.

3. (optional) A spontaneous thought when seeing the new storybook layout, is that users might get confused if they only see the "expandable rows" story, meaning they might not figure out that they can scroll to see all other stories. Would it be a good idea to add some text hinting that there are more stories below? (i have a small screen, maybe that is the problem)

theJohnnyMe:
@alexlioliopoulos  - Things you mentioned are fixed and visible now on the same link.

During work I noticed I need to fix:

1.  table title styling for HTML&CSS version
2. fix the compact size and vertical dividers for expandable rows
3. add some more controls to specific stories

Please let me know if you notice anything else ;)

alexlioliopoulos:
First very small thing I noticed, is that all tables come with a table header, except for the basic and the exp rows one.

theJohnnyMe:
@alexlioliopoulos  - Please check now what stories look like. I have removed the "title" area for those versions that do not need it anymore. I have added a story for custom width too.

alexlioliopoulos:
Super small comment: the min width story does not have numbers in the last col

image

alexlioliopoulos:
Looks amazing MrJ, well done!

albinscania (approved):
A lot to take in, but code looks good! Nothing to comment on there!

ElviraElmira:
Hello! I have tested the table component for some scenarios in an upcoming app! Here are some questions/issues I have had:

- The ability to set a full width on the whole table would be nice, with evenly distributed column widths (table layout fixed) We don't want to set fixed widths on the columns, and the basic layout becomes to wide with our datasets. no-min-width can work, but:
- I tried to implement custom filtering, not using the builtin filter input, but from "outside". When I have dynamic data for the rows, that changes by my filtering function, the table grows in width and expands all containers. Might have to do with the rerendering in the Angular change detection... but it seems it's losing the 'no-min-width' setting. Not sure what I'm doing wrong :) 
- We will need editable cells, so basically putting text-fields and selects into cells. It might be needed to decrease the paddings in the cells in those cases. Or what does the design say about it? :)
- Also, responsive mode - what is the pattern for the responsiveness? Can displaying the table as "cards" be a possibility?
- Sorting did not work when I had custom html in body-cells, but I might need to implement a custom sorting function, right?

theJohnnyMe:
> 

@ElviraElmira Thanks a lot for the valuable feedback! 

I think that the issue described in the first two bullet points you brought up is tiny bugs. So what happens is that <table> tag sends out the no-min-width rule to all sub-components, but if sub-components have to rerender, which happens, maybe they lose that setting and the parent component is not sending it over again. I will try to debug it myself and come up with some feedback. 

Responsive part - I was testing myself, how I managed to get the table responsive by using custom-width and no-min-width. For example, if I had 4 columns and I wanted to make it fully responsive I added custom-width="25vw"  - aka 1/4th of 100vw. It is tricky to make a simple solution for it as to have all things covered user would have to set min-width, max-width and width rule for every column but maybe that is the way to go? 

Regarding padding in the data-table, I think it is more on design side. What do you say @alexlioliopoulos ?

Sorting and additional HTML content in cell - not supported by our side.

"No data" row - so for now it is basically a single row with cell that has attribute of "colSpan" value ste tot he number of the columns - then it long as long are other rows. I can show you how you can code this - it is rather simple.

Question: Do you have an upper code in some branch of some repo I can clone and try it out?

alexlioliopoulos:
Regarding the cell editing, I think it's an interesting one to solve. We have no design for such a case yet sadly, however I'd say that it makes sense to just replace the current cell with the medium variant of our text input. They should be both 48px tall, so it would be seemless. The width should be the one of the cell. Does this make sense?

ElviraElmira:
Hey! Sorry for the late reply, it seems the notifications from this came to my gmail and I totally missed it. The input suggestion looks good @alexlioliopoulos. Regarding the loosing of dimension, I have no current branch to test this out, I had a while ago while testing, but it's gone now. But what I basically did is having the data creating rows with *ngFor and then a filtering function. So while filtering non matching rows were removed from the dom and when erasin the filter rows were put back and when put back the table grew.

EmelieLitwin:
Amazing job! 
Just some random thoughts after going through storybook today:

- Not sure about what the correct behavior is, but if you choose to make it sortable the arrows don't show up until you click the column.
- Would be nice to minimize the number of controls to make it a little more user-friendly. Maybe use a boolean toggle to make more choices available. For example, a control called "Sortable" and if you make it true "Column 1 is sortable", "Column 2 is sortable", etc becomes visible. The sortable choices could be a checkbox list maybe. Generally, radio buttons are better than dropdown when there are few choices and text-align only have two choices, but it would also make the control list longer

Again, such a great job! 😍

theJohnnyMe:
I have rebased this one with latest main, so use this version with theme-light: 4.4.0.

Also fix for counting expandable rows is added. 

Let's talk about this on Monday @ElviraElmira. I have one more colleague reporting similar issue like you did.

PR #422, opened 2022-08-24
feat: Slider component

albinscania:
**Describe pull-request**  
Created SLIDER component according to design here:
https://www.figma.com/file/yOgycQh7fsEtBz77CyUtRk/Slider?node-id=69%3A7716

NOTE: Dark Theme is not yet implemented.

Fixes: #1903

**How to test**  
Checkout this branch and navigate to the slider-component. The old ALPHA-one is still there for legacy reasons but we can probably remove that this will be merged.

* divider was changed to "ticks" to use the same terminology as the figma
* bug fix for initial value (scrubber position) when min is > 0
* snap-attribute added to make scrubber snap to ticks.
*


alexlioliopoulos:
I will check it again tomorrow more thoroughly, but it seems to be splendid!

I only have a weird comment: it seems that for some reason you need to scroll to see the slider component in storybook

image

components/src/components/slider/slider.scss
```diff
@@ -0,0 +1,338 @@
...
+  .sdds-slider__controls {
+    position: relative;
+    top: -25px;
+
+    .sdds-slider__control {
+      cursor: pointer;
+
+      &.sdds-slider__control-minus {
+        padding: 18px;
+        padding-left: 0;
```
theJohnnyMe, line 45: To keep it neat, maybe it is better to use padding: 18px 18px 18px 0; ? 
Same goes for line 48.

components/src/components/slider/slider.scss
```diff
@@ -0,0 +1,338 @@
...
+
+      input.sdds-slider__input-field {
+        @include type-style("detail-02");
+
+        color: var(--sdds-grey-700);
+        border: 0;
+        background-color: transparent;
+        text-align: center;
+        padding: 12px;
+        box-shadow: inset 0 -1px 0 #cdd1db;
```
theJohnnyMe, line 84: Do we have these colors as CSS variables?
The same goes for lines 88 and 92. 
Please check.

components/src/components/slider/slider.scss
```diff
@@ -0,0 +1,338 @@
...
+        }
+      }
+
+      &.pressed {
+        width: 24px;
+        height: 24px;
+
+        &::before {
+          display: block;
+          opacity: 0.16 !important;
```
theJohnnyMe, line 180: Justo noticed this one. Is there a way to avoid usage of important here or it complicates stuff too much?

components/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,446 @@
...
+  useInput: boolean = false;
+  useSmall: boolean = false;
+
+  @Prop() label = '';
+
+  @Prop() value = '0';
+
+  @Prop() min = '0';
+
+  @Prop() max = '100';
```
theJohnnyMe, line 34: Please provide a comment above each prop on what it does in a form /** Test 123 */ - it ends up in an autogenerated readme which is super useful for other developers.

components/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,446 @@
...
+    this.controlsStep(parseInt(this.step));
+  }
+
+  componentWillLoad() {
+    const numDividers = parseInt(this.dividers);
+
+    if (numDividers > 0) {
+      this.dividerValues = [this.getMin()];
+
+      /*const step =
```
theJohnnyMe, line 253: I guess this can be removed.

components/src/alpha/slider/slider.tsx
```diff
@@ -1,11 +1,11 @@
 import { Component, h, State, Listen, Prop, Watch } from '@stencil/core';
 
 @Component({
```
theJohnnyMe, line 3: Just delete this entire component folder and files, it is not needed anymore. 
Glad that you are careful though.

theJohnnyMe (changes requested):

components/src/components/slider/slider.stories.js
```diff
@@ -0,0 +1,184 @@
+const style = `
+  <style>
+    .storybook-slider-wrapper {
```
theJohnnyMe, line 3: This one gives false positives in visual testing. A component might need some padding-top / margin-top to display correctly. Let me show you in devCheckIn.

components/src/components/slider/slider.stories.js
```diff
@@ -0,0 +1,184 @@
...
+    },
+    showTooltip: {
+      name: 'Show tooltip',
+      description: 'Show or hide tooltip',
+      control: {
+        type: 'boolean',
+      },
+    },
+    showDividers: {
+      name: 'Show dividers',
```
theJohnnyMe, line 67: Question: Should we count in start & end dividers too? When you have 3 as the value, it is actually 5. Just to avoid confusion.

theJohnnyMe: Or should we simply say how many sections you want, meaning if you want 0,10,20,30... to 100, you need to say I want 10 sections. Just thinking was it easier to understand for the end user?

components/src/components/slider/slider.stories.js
```diff
@@ -0,0 +1,184 @@
...
+    },
+    disabled: {
+      name: 'Disabled',
+      description: 'Put control in disabled state',
+      control: {
+        type: 'boolean',
+      },
+    },
+    small: {
+      name: 'Small',
```
theJohnnyMe, line 124: Maybe name it "Smaller thumb" or "Small indicator" as the only the size of that particle changes, not an entire component.

theJohnnyMe (changes requested):

components/src/components/slider/slider.scss
```diff
@@ -0,0 +1,338 @@
...
+      transform: translateX(-50%);
+      top: 34px;
+    }
+  }
+
+  .sdds-slider__scrubber {
+    position: absolute;
+
+    .sdds-slider__scrubber-inner {
+      /* width: 16px;
```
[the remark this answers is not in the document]
albinscania: Removed all comments!

components/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,446 @@
...
+  scrubberGrabPos: object = { x: 0, y: 0 };
+  scrubberGrabbed: boolean = false;
+  scrubberLeft: number = 0;
+  dividerValues: Array<number> = [];
+  disabledState: boolean = false;
+  useControls: boolean = false;
+  useInput: boolean = false;
+  useSmall: boolean = false;
+
+  @Prop() label = '';
```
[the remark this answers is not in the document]
albinscania: Added prop descriptions!

components/src/components/slider/slider.tsx
```diff
@@ -0,0 +1,446 @@
+import { Component, Prop, Listen, Host, h } from '@stencil/core';
+
+@Component({
+  tag: 'sdds-slider',
+  styleUrl: 'slider.scss',
+  shadow: true,
+})
+export class Slider {
+  wrapperElement: HTMLElement = null;
```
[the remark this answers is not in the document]
albinscania: Added @element() decorator for conformity!

components/src/components/slider/slider.scss
albinscania: @theJohnnyMe Yes we did! Updated!

components/src/components/slider/slider.scss
albinscania: I think it gets slightly messier to move that block of css (which would be needed), into the nested spot.
It's the only !important in the whole .scss though ;)

components/src/components/slider/slider.tsx
albinscania: Fixed

components/src/components/slider/slider.tsx
albinscania: It mostly def. can! fixed.

components/src/alpha/slider/slider.tsx
albinscania: Deleted!

components/src/components/slider/slider.stories.js
albinscania: I updated the container (still have some style there though, but it makes more sense).
Also added a sensible default top-padding to the component itself like we discussed.

components/src/components/slider/slider.stories.js
albinscania: I think it kind of makes sense the way it is now (not counting start and end). Also, if a user gets it wrong the first time it should be kind of clear what they need to to imho.

I'll hold on to this for a while until we talked about renaming them and stuff.

components/src/components/slider/slider.stories.js
albinscania: I'll let Matylda or Alex decide this. Works for me to call the prop "small-thumb" instead if we want.

components/src/components/slider/slider.tsx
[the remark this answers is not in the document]
theJohnnyMe: Awesome! ;)

components/src/components/slider/slider.scss
theJohnnyMe: Great!

components/src/components/slider/slider.scss
theJohnnyMe: Sure, no problem, it is just to double check ;)

components/src/components/slider/slider.stories.js
theJohnnyMe: Great!

theJohnnyMe:
Check if snapping can get of divider, for example in this one: 25, 50, 75, 100...
Screenshot 2022-08-30 at 08 57 09

EmelieLitwin:
Amazing job!
I tried the "snapping" and it snaps to 24, 49, 74 etc for me (or at least that's what the tooltip say)

EmelieLitwin (approved):
Looks good to me 😍

theJohnnyMe (approved):
Works like a charm!

Lets rebase this one together before merge to resolve old code jumping in or merge conflicts :)

PR #423, opened 2022-08-25
feat: popover-canvas component

albinscania:
**Describe pull-request**  
A first version of the popover canvas component.

**Solving issue**  
Fixes: #2042

**How to test**  
Checkout the branch and navigate to the popover component.
Click the little "kebab"-icon to trigger the canvas popup, which has some sample content.

Now with updated z-index to be correctly placed.


alexlioliopoulos (approved):
Another great one, thanks Albin! 

Just a question: is there a difference between bottom and bottom-start? They seem to be giving the exact same placement for me.

albinscania:
@alexlioliopoulos yeah, that's a good point. I used the same ones from "tooltip", but I also can't really see the difference here. I can clean it up a bit!

components/src/components/popoverCanvas/popovercanvas.tsx
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    this.popoverCanvasElement.addEventListener('mousedown', (event) => {
+      event.stopPropagation();
+    });
+  }
+
+  render() {
+    return (
+      <div
```
theJohnnyMe, line 89: This is a great candidate to use Host tag instead as then you do not need to use reference.

components/src/components/popoverCanvas/popovercanvas.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+@Component({
+  tag: 'sdds-popover-canvas',
+  styleUrl: 'popovercanvas.scss',
+  shadow: true,
+})
+export class PopoverCanvas {
+  popoverCanvasElement!: HTMLElement;
+
+  /** The CSS-selector that will trigger this Popover Canvas */
```
theJohnnyMe, line 13: Great job with adding comments ;)

components/src/components/popoverCanvas/popovercanvas.tsx
```diff
@@ -0,0 +1,99 @@
+import { Component, Listen, h, Prop, State } from '@stencil/core';
+import { createPopper } from '@popperjs/core';
+import type { Placement } from '@popperjs/core';
+
+@Component({
+  tag: 'sdds-popover-canvas',
```
theJohnnyMe, line 6: Not aimed for this line, but for naming files and folder. I would suggest renaming the folder and files to one that uses the hyphen: popover-canvas, popover-canvas.scss... Just to keep consistent with side-menu, code-sample, radio-button etc...

theJohnnyMe (changes requested):

theJohnnyMe (approved):
Looks good to me.

PR #424, opened 2022-08-26
feat: popover-menu component 

albinscania:
**Describe pull-request**  
A new component in the form of a Popover Menu, based on core-code from the popover canvas component.

**Solving issue**  
Fixes: #2043

**How to test**  
Go to the netlify link or checkout the branch, and browse to the PopoverMenu component. Click the kebab-dots to see it in action, both with and without icon.

Now with updated z-index to be correctly placed.

### DIFF
--- a/components/src/components/popover-menu/documentation.txt
+++ b/components/src/components/popover-menu/documentation.txt
@@ -1,0 +1,108 @@
+The popover menu is a component made for context menus.
+It requires a valid selector as an attribute and then it will be automatically opened when the user clicks on the element with this selector.
+
+There are two versions, one with and one without icons. Both variants supports a placement-attribute to control where the menu should popup when the element is clicked.
+
+Valid values are:
+
+'auto', 'auto-left', 'auto-right', 'bottom', 'bottom-start', 'bottom-end', 'top', 'top-start', 'top-end', 'left', 'left-start', 'left-end', 'right', 'right-start, 'right-end'
+
+#Popover menu
+
+<sdds-popover-menu
+    placement="auto"
+    selector="#trigger"> 
+    <ul class="sdds-popover-menu-wrapper">
+        <li>
+            <a target="_blank" href="https://digitaldesign.scania.com">Menu item 1</a>
+        </li>
+        <li>            
+            <a target="_blank" href="https://digitaldesign.scania.com">Menu item 2</a>        
+        </li>
+        <li>
+            <a target="_blank" href="https://digitaldesign.scania.com">Menu item 3</a>
+        </li>
+        <li class="divider"></li>
+        <li>        
+            <a target="_blank" href="https://digitaldesign.scania.com">Menu item 4</a>
+        </li>
+        <li>
+            <a target="_blank" href="https://digitaldesign.scania.com">Menu item 5</a>
+        </li>
+    </ul>        
+</sdds-popover-menu>
+
+<button id="trigger">Click to open popup menu</div>
+
+
+
+#Popover menu with icons
+
+<sdds-popover-menu
+    placement="auto"
+    selector="#trigger2"> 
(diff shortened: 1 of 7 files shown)


alexlioliopoulos:
It is grey-100! 

(sdds-library in Figma is usually the source of truth when in doubt)
image

alexlioliopoulos:
Looks great as Matylda said! 

It is a tradition though to have at least one question in every PR so I won't break it here :D 

It might not have been clear on the task itself, but the idea was that either have icons for all items, or for none of them. Could we set it to be like that? Also we want devs using this to be able to add a divider wherever they want. Is it flexible like that?

albinscania:
@alexlioliopoulos Ah I see... I guess I can add something like an "with-icon" class and then just indent the text where they choose not to add an icon? (even if that's against the design).

And yes, dividers can be added anywhere :)

alexlioliopoulos:
> 

That would be an even greater solution - let's go with it if it is easy to make!

components/src/components/popoverMenu/popovermenu.scss
```diff
@@ -0,0 +1,94 @@
...
+
+    &:last-child {
+      margin-bottom: 0;
+    }
+
+    a {
+      text-decoration: none;
+      position: relative;
+      width: 100%;
+      display: block;
```
theJohnnyMe, line 38: As one on 40 is flex, this one can be removed.

components/src/components/popoverMenu/popovermenu.scss
```diff
@@ -0,0 +1,94 @@
+@import "../../helpers/style-helpers.scss";
+@import "@scania/typography/dist/scss/mixins";
+@import "@scania/typography/dist/scss/tokens";
+
+.sdds-popover-menu {
+  overflow: hidden;
+  width: 160px;
+  background-color: #fff;
```
theJohnnyMe, line 8: Not a biggy but could be var(--sdds-white)...

components/src/components/popoverMenu/popovermenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    this.popoverMenuElement.addEventListener('mousedown', (event) => {
+      event.stopPropagation();
+    });
+  }
+
+  render() {
+    return (
+      <div
+        ref={(el) => (this.popoverMenuElement = el as HTMLElement)}
```
theJohnnyMe, line 88: As in popover canvas, this is a good chance to use Host :)

components/src/components/popoverMenu/popovermenu.tsx
```diff
@@ -0,0 +1,95 @@
+import { Component, Listen, h, Prop, State } from '@stencil/core';
+import { createPopper } from '@popperjs/core';
+import type { Placement } from '@popperjs/core';
+
+@Component({
+  tag: 'sdds-popover-menu',
+  styleUrl: 'popovermenu.scss',
```
theJohnnyMe, line 7: as in popover canvas, see if folder name and files in can be renamed to popover-menu, popover-menu.scss etc... just to align with rest of "two-words" components...

components/src/components/popoverMenu/popovermenu.stories.js
```diff
@@ -0,0 +1,212 @@
...
+          'right',
+          'right-start',
+          'right-end',
+        ],
+      },
+      defaultValue: 'right-start',
+      description: 'Position of the PopoverMenu',
+    },
+    withIcons: {
+      name: 'withIcons',
```
theJohnnyMe, line 29: In the version of the story, I checked there were no icons showing when this one was true. Please check.

theJohnnyMe (changes requested):

theJohnnyMe (approved):
Looks good to me.

Issue #265, opened 2021-12-23
Improvement - Missing empty badge version


mmexvr:
**Describe current situation**  
_Describe what the current situation and what problem it is causing for this improvement_
Missing badge without a number, simple dot. See screenshot

**Describe the improvement**  
_A clear and concise description of what kind of improvement you have in mind._

**If applicable, please provide steps to reproduce**
_Steps to reproduce the behavior:_  
1. Go to badge component in storybook or website
2. No badge component without a number exist 

**Screenshots**  
_If applicable, add screenshots to help explain your ideas._
<img width="348" alt="Screenshot 2021-12-23 at 15 06 42" src="https://user-images.githubusercontent.com/35451568/147251428-ec7d510b-a33e-4332-942c-939e803eece4.png">

**Additional context**  
_Add any other context about the suggestion here._

alexlioliopoulos:
It seems that the empty badge variant exists, it is just not documented in the website and storybook. However it works if you give no value to it. We are currently working on the documentation.

PR #429, opened 2022-08-29
Empty badge version

EmelieLitwin:
**Describe pull-request**  
Added a prop for small size (simple red dot without number) and one for visibility.

**Solving issue**  
Github: [#265](https://github.com/scania-digital-design-system/sdds/issues/265)
Azure: [AB#1337](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1337)

### DIFF
--- a/components/src/components/badges/badges-component.tsx
+++ b/components/src/components/badges/badges-component.tsx
@@ -6,13 +6,13 @@ import { Component, h, Prop, Watch, State } from '@stencil/core';
   shadow: true,
 })
 export class SddsBadges {
-  /**Value shown in badge**/
+  /**Value shown in badge*/
   @Prop() value: string = '';
 
-  /**Changes visibility of badge**/
+  /**Changes visibility of badge*/
   @Prop() isVisible: boolean = true;
 
-  /**Changes badge from default to small size**/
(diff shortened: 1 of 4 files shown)


albinscania (approved):
Looks good! Nothing that you added that I think needs to change. (there is this comment about parsing the number, but that was there from before, but I would remove it in the future :)

alexlioliopoulos:
Nice! My only comment is regarding the fact that we have the value field in the small variant.

image

which is not applicable in that story, right?

components/src/components/badges/badges-component.tsx
```diff
@@ -6,43 +6,85 @@ import { Component, h, Prop, Watch, State } from '@stencil/core';
   shadow: true,
 })
 export class SddsBadges {
+  /**Value shown in badge*/
   @Prop() value: string = '';
 
+  /**Changes visibility of badge*/
```
theJohnnyMe, line 12: Super nice with using comments ;)

theJohnnyMe (approved):
NIce work Emelie ;)

PR #433, opened 2022-09-01
fix: defining sizing props

ahujamoh:
**Describe pull-request**  
I found a few of the type were string, they should be only accepting a few strings so rest

**Solving issue**  
_No issues getting directly, but saving future issues which can be undefined behaviour_

**How to test**  
1. Go to storyboard and try out the components such as badge, datetime picker
2. The changes limit to a few changes only

**Screenshots**  
No screenshots

### DIFF
--- a/components/src/components/badges/badges-component.tsx
+++ b/components/src/components/badges/badges-component.tsx
@@ -17,7 +17,7 @@ export class SddsBadges {
 
   @State() shape: string = '';
 
-  @Prop() size: 'sm' | 'md' | '' = 'sm';
+  @State() size: string = 'sm';
 
   @State() text: string = '';
 
--- a/components/src/components/badges/badges-component.tsx
+++ b/components/src/components/badges/badges-component.tsx
@@ -17,7 +17,7 @@ export class SddsBadges {
 
   @State() shape: string = '';
 
-  @State() size: string = 'sm';
+  @Prop() size: 'sm' | 'md' | '' = 'sm';
 
   @State() text: string = '';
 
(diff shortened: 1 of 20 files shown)


components/src/components/badges/badges-component.tsx
```diff
@@ -17,7 +17,7 @@ export class SddsBadges {
 
   @State() shape: string = '';
 
-  @State() size: string = 'sm';
+  @Prop() size: 'sm' | 'md' | '' = 'sm';
```
theJohnnyMe, line 20: This one is a bit tricky. So Badge as a component does not have a size prop as it does not have sizes really. It has two states better to say. One state is when we want to display a number in it and the second one is where we want to display just a red dot as an indicator without any value inside it. I would recommend reverting this change.

As you @EmelieLitwin worked recently on this one, what are your thoughts?

theJohnnyMe:
@ahujamoh - Thanks a lot buddy for your contribution. Really improvement when it comes to code hygiene and more defined props. Nice to try to add unit test too for accordion, it is also our aim to add more of it in the future as well. 
We are really happy to have more and more of our colleagues involved in improvements of the Tegel design system :)

components/src/components/badges/badges-component.tsx
EmelieLitwin: I think I agree with the recommendation to revert this back too.

But otherwise: thank you so much for your contribution, great job! 😍 💪

ahujamoh: I have reverted this one

theJohnnyMe (approved):
Looks good to me!

theJohnnyMe:
@ahujamoh  - Thanks again for really nice work! It is merged and will go out with next release ;)

PR #436, opened 2022-09-02
feat: new icons in icon package

theJohnnyMe:
**Describe pull-request**  
Upgrading icons package to support new icons

**Solving issue**  
Fixes: AB#2050

**How to test**  
1. Go to Storybook
2. Open Icons -> Component
3. Try different icons from the dropdown
4. Try changing sizes

**Additional context**  
UNDER DEVELOPMENT !


alexlioliopoulos:
One initial comment, is that the size selection is in rem, making the icons that are rendered super large. Also, there are steps of 4rems, meaning that there are only the 16px and the 32px icons rendered, whereas we have the 20 and 24px variants normally.

theJohnnyMe:
> One initial comment, is that the size selection is in rem, making the icons that are rendered super large. Also, there are steps of 4rems, meaning that there are only the 16px and the 32px icons rendered, whereas we have the 20 and 24px variants normally.

This one is replaced with "px" values starting at 16 an going to 100, with increments of 4.

theJohnnyMe:
@alexlioliopoulos and others reviewing this one - I also did some changes to the button component. Let's discuss it more next week.

alexlioliopoulos (approved):

matyldamc (approved):

theJohnnyMe:
@alexlioliopoulos, @charlottelorensson and @matyldamc - Please check the Icons component and Webfont solution, but also check the button with icons too. I did some fixes there so it should finally look good :)

alexlioliopoulos:
The first picture is showing that the text in the button is not exactly centered. In fact, it might be centered to the small letters. (values in px are indicative, they are very large because of zoom in)

The second image is the coded components vs the designed one, on on top of the other.

image

image

alexlioliopoulos:
Apart from the comment above, which is not a big problem anyway, icons seem to work perfect. Except for the Pin icon of course, which however is on me to find a way to fix.

matyldamc:
@alexlioliopoulos but isn't it because it needs to accommodate for descenders (add a lower case Y to the word to see what I mean).

alexlioliopoulos:
> @alexlioliopoulos but isn't it because it needs to accommodate for descenders (add a lower case Y to the word to see what I mean).

Makes sense! However it is still the case that on design side it is placed differently.

image

albinscania (approved):
Great with the added comments and the cleanup.

PR #437, opened 2022-09-07
refactor(badges): Simplified code and added size prop

EmelieLitwin:
**Describe pull-request**  
Simplified the badges code, removing some States and adding a size prop to be more consistent with other components

### DIFF
--- a/components/src/components/badges/badges-component.tsx
+++ b/components/src/components/badges/badges-component.tsx
@@ -12,12 +12,13 @@ export class SddsBadges {
   /**Changes visibility of badge*/
   @Prop() isVisible: boolean = true;
 
-  /**Changes badge from default to small size*/
+  /** !!Deprecated!! Use size prop instead. Changes badge from default to small size*/
   @Prop() isSmall: boolean = false;
 
-  @State() shape: string = '';
-
+  /**Component is avaible in size default and small (small dot). Default size is default*/
   @Prop() size: string = 'default';
+
+  @State() shape: string = '';
 
   @State() text: string = '';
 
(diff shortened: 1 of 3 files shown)


alexlioliopoulos (approved):

components/src/components/badges/badges-component.tsx
```diff
@@ -17,70 +17,59 @@ export class SddsBadges {
 
   @State() shape: string = '';
 
-  @State() size: string = 'sm';
+  @Prop() size: string = 'default';
```
theJohnnyMe, line 20: Do you think it would be possible to specify usage of only 'small' OR 'default' option as in [this PR](https://github.com/scania-digital-design-system/sdds/pull/433/files)?

components/src/components/badges/badges-component.tsx
```diff
@@ -17,70 +17,59 @@ export class SddsBadges {
...
   @Watch('value')
   @Watch('isVisible')
   @Watch('isSmall')
+  @Watch('size')
   watchProps() {
     this.checkProps();
   }
   componentWillLoad() {
     this.checkProps();
+    if (this.isSmall) {
```
theJohnnyMe, line 33: Nice fallback!

components/src/components/badges/badges-component.tsx
```diff
@@ -17,70 +17,59 @@ export class SddsBadges {
...
     } else {
-      this.defaultVariation = false;
-      this.smallVariation = false;
+      this.value !== '' && this.size !== 'sm'
+        ? console.warn(
+            'The provided value is either empty or string, please provide value as number.'
+          )
+        : '';
     }
+
```
theJohnnyMe, line 53: I guess it can be deleted if not needed.

components/src/components/badges/badges-component.tsx
```diff
@@ -17,70 +17,59 @@ export class SddsBadges {
...
+    // }
   }
 
   render() {
     return (
       <host>
-        {this.defaultVariation && (
+        {this.size === 'default' && this.isVisible && (
           <div
             class={`sdds-badges sdds-badges-${this.size} sdds-badges-${this.shape}`}
```
theJohnnyMe, line 67: If we have a conditional logic to render based on size, maybe we can remove `sdds-badges-${this.size}` be removed from the class?

components/src/components/badges/badges-component.tsx
```diff
@@ -17,70 +17,59 @@ export class SddsBadges {
...
           >
             <div class="sdds-badges-text">{this.text}</div>
           </div>
         )}
-        {this.smallVariation && (
+        {this.size === 'sm' && this.isVisible && (
```
theJohnnyMe, line 72: Maybe we can use visibility: hidden, display: none in this case, or even opacity: 0, rather than add/remove a piece of code from DOM. Let's have a quick catch-up on it.

theJohnnyMe (approved):
Nice work Emelie! ;)

PR #439, opened 2022-09-08
feat(dropdown): new multiselect type

EmelieLitwin:
**Describe pull-request**  
A new multiselect variation of the dropdown component

**Solving issue**  
[AB#506](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/506) Build component - Multiselect

**How to test**  
1. Go to storybook and test functionality

### DIFF
--- a/components/src/components/dropdown/dropdown-filter.tsx
+++ b/components/src/components/dropdown/dropdown-filter.tsx
@@ -40,8 +40,8 @@ export class DropdownFilter {
   /** Add the value of the option to set it as default */
   @Prop() disabled: boolean;
 
-  /** `large` (default), `small`, `medium` */
-  @Prop() size: 'small' | 'medium' | 'large' = 'large';
+  /** Controls the size of dropdown. 'sm', 'md' and 'lg' correct values and 'small', 'medium' and 'large' are deprecated */
+  @Prop() size: 'sm' | 'md' | 'lg' | 'small' | 'medium' | 'large' = 'lg';
 
   /** Set to true to make the width following the label text length */
   @Prop() inline: boolean = false;
--- a/components/src/components/dropdown/dropdown-option.tsx
+++ b/components/src/components/dropdown/dropdown-option.tsx
@@ -54,7 +54,10 @@ export class DropdownOption {
       });
     }
   }
-
+  @Listen('click')
+  handleClick(ev) {
+    ev.stopPropagation();
+  }
   componentWillLoad() {
     this.innerValue = this.value;
     this.isMultiSelectOption = this.host
--- a/components/src/components/dropdown/dropdown-option.tsx
+++ b/components/src/components/dropdown/dropdown-option.tsx
@@ -34,10 +34,15 @@ export class DropdownOption {
     bubbles: true,
   })
   selectOption: EventEmitter<any>;
+  isMultiSelectOption: boolean;
 
   @Listen('mouseover')
   changeFocusHandler() {
     this.host.focus();
+  }
+  @Listen('mouseout')
+  removeFocusHandler() {
+    this.host.blur();
   }
   @Listen('keydown')
   onKeyDown(event: KeyboardEvent) {
@@ -52,16 +57,31 @@ export class DropdownOption {
 
   componentWillLoad() {
     this.innerValue = this.value;
+    this.isMultiSelectOption = this.host
+      .closest('sdds-dropdown')
+      .classList.contains('sdds-dropdown-multiselect');
   }
 
   selectOptionHandler(value) {
     const listOptions = value.parent.childNodes;
     this.selectOption.emit(value);
-
-    listOptions.forEach((optionEl) => {
-      optionEl.selected = false;
-    });
-    this.selected = true;
+    if (!this.isMultiSelectOption) {
+      listOptions.forEach((optionEl) => {
+        optionEl.selected = false;
+      });
+    }
+    var optionCheckbox = this.host.shadowRoot.querySelector('input');
+    if (this.selected) {
+      this.selected = false;
+      if (optionCheckbox) {
+        optionCheckbox.checked = false;
+      }
+    } else {
+      this.selected = true;
+      if (optionCheckbox) {
+        optionCheckbox.checked = true;
+      }
+    }
   }
 
   render() {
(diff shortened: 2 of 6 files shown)


alexlioliopoulos (changes requested):
Amazing work! I only have one small comment:

image

I think there is a weird white line on the left side of the component, which only shows when a selection has been made. Would it be possible to remove it?

Update:

We are also missing 1) the two label variants, and 2) the read-only version.

As stated in our chat, it is not a hard requirement for this sprint. We can push something out now, and add the extra things next time.

alexlioliopoulos:
> Amazing work! I only have one small comment:
> 
> image
> 
> I think there is a weird white line on the left side of the component, which only shows when a selection has been made. Would it be possible to remove it?
> 
> **Update:**
> 
> We are also missing 1) **the two label variants**, and 2) **the read-only version**.
> 
> As stated in our chat, it is not a hard requirement for this sprint. We can push something out now, and add the extra things next time.

I played around a bit and it seems that the white stripe appears only in some zoom levels. 125, 90, 75 are where the issue happens; in all other zoom levels it does not show up.

components/src/components/dropdown/dropdown-filter.tsx
```diff
@@ -40,8 +40,8 @@ export class DropdownFilter {
   /** Add the value of the option to set it as default */
   @Prop() disabled: boolean;
 
-  /** `large` (default), `small`, `medium` */
-  @Prop() size: 'small' | 'medium' | 'large' = 'large';
+  /** `lg` (default), `sm`, `md` */
+  @Prop() size: string = 'lg';
```
theJohnnyMe, line 44: Nice!

ahujamoh, line 44: it could be 

@Prop() size: 'sm' | 'md' | 'lg' = 'lg';

keeping the size property strongly typed (as we know these will be the expected types)

components/src/components/dropdown/dropdown.tsx
```diff
@@ -35,8 +35,8 @@ export class Dropdown {
   /** `default`, `multiselect`, `filter`, `nested` */
   @Prop() type: 'default' | 'multiselect' | 'filter' | 'nested' = 'default';
 
-  /** `large` (default), `small`, `medium` */
-  @Prop() size: 'small' | 'medium' | 'large' = 'large';
+  /** `lg` (default), `sm`, `md` */
+  @Prop() size: string = 'lg';
```
ahujamoh, line 39: (same comment as before)

it could be

@prop() size: 'sm' | 'md' | 'lg' = 'lg';

keeping the size property strongly typed (as we know these will be the expected types)

components/src/components/dropdown/dropdown.tsx
```diff
@@ -80,15 +86,34 @@ export class Dropdown {
   componentWillLoad() {
     // If default option is set, update the default selectedLabel value
     // this.host.children is a HTMLCollection type, cannot use forEach
-    this.setOptionFromOutside(this.defaultOption);
     this.listItemArray = Array.from(this.host.children);
-  }
+    for (let i = 0; i < this.listItemArray.length; i++) {
+      this.optionValues.push(this.listItemArray[i].value);
```
ahujamoh, line 91: We can use imperative style instead of being declarative, would make it easy to read.

```
this.listItemArray.map((listItem) => {
  this.optionValues.push(listItem.value);
  this.optionLabels.push(listItem.label);
});
```

components/src/components/dropdown/dropdown-filter.tsx
EmelieLitwin: Good point! Thank you for the feedback Mohit! 😄

components/src/components/dropdown/dropdown.tsx
```diff
@@ -121,6 +149,10 @@ export class Dropdown {
       this.open = false;
     }
   }
+  @Listen('optionClicked', { target: 'document' })
+  closeMultiselect() {
+    this.open ? this.open : !this.open;
```
ahujamoh, line 154: may be i am wrong, but shouldn't this be assigned to `this.open`?

ahujamoh (commented):
made some suggestion

components/src/components/dropdown/dropdown.tsx
```diff
@@ -80,15 +86,34 @@ export class Dropdown {
...
-    this.setOptionFromOutside(this.defaultOption);
     this.listItemArray = Array.from(this.host.children);
-  }
+    for (let i = 0; i < this.listItemArray.length; i++) {
+      this.optionValues.push(this.listItemArray[i].value);
+      this.optionLabels.push(this.listItemArray[i].innerText);
+    }
+    this.setOptionFromOutside(this.defaultOption);
 
+    if (this.size === 'small') {
```
theJohnnyMe, line 96: Nice!

alexlioliopoulos (changes requested):
Something seems to be off now, it was not like that before. Check here:

image

Also, not sure if it was like that before, but at least now when you select an option, the dropdown collapses and if you want to select more, you need to expand it again. Should it be like that?

EmelieLitwin:
> Something seems to be off now, it was not like that before. Check here:
> 
> image
> 
> Also, not sure if it was like that before, but at least now when you select an option, the dropdown collapses and if you want to select more, you need to expand it again. Should it be like that?

Thank you for your fast reviews @alexlioliopoulos! Something happened in the rebase but I hope it's fixed now. :)

alexlioliopoulos:
> > Something seems to be off now, it was not like that before. Check here:
> > image
> > Also, not sure if it was like that before, but at least now when you select an option, the dropdown collapses and if you want to select more, you need to expand it again. Should it be like that?
> 
> Thank you for your fast reviews @alexlioliopoulos! Something happened in the rebase but I hope it's fixed now. :)

They are both fixed now, super nice!

theJohnnyMe (approved):
Nice work Emelie! It looks good to me :)

components/src/components/dropdown/dropdown-option.tsx
```diff
@@ -49,19 +54,37 @@ export class DropdownOption {
       });
     }
   }
-
+  @Listen('click')
+  handleClick(ev) {
+    ev.stopPropagation();
+  }
```
mJarsater, line 60: This will stop all custom onClick

PR #441, opened 2022-09-14
feat(header): search-bar

victorscania:
**Describe pull-request**  
Makes the searchbar collapse in to the side menu on lower breakpoints

**Solving issue**  
Fixes: [AB#1733](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1733)

**How to test**  
Check the storybook link (Header > Searchbar menu). go to a breakpoint below the threshold and test around.

### DIFF
--- a/components/src/components/header/header.scss
+++ b/components/src/components/header/header.scss
@@ -1,7 +1,7 @@
-@import '../../../node_modules/@scania/typography/dist/scss/mixins';
-@import '../../../node_modules/@scania/typography/dist/scss/tokens';
-@import '../../../node_modules/@scania/grid/dist/scss/grid';
-@import '../../helpers/components-shared';
+@import "../../../node_modules/@scania/typography/dist/scss/mixins";
+@import "../../../node_modules/@scania/typography/dist/scss/tokens";
+@import "../../../node_modules/@scania/grid/dist/scss/grid";
+@import "../../helpers/components-shared";
 
 @mixin sdds-nav__basic-element-style {
   cursor: pointer;
@@ -35,7 +35,7 @@
   justify-content: space-between;
 
   .sdds-nav__overlay {
-    content: '';
+    content: "";
     width: 100%;
     height: 100%;
     position: fixed;
@@ -146,7 +146,8 @@
 
     display: block;
     cursor: pointer;
-    background-image: var(--sdds-background-image-scania-symbol-svg),
+    background-image:
+ var(--sdds-background-image-scania-symbol-svg),
       var(--sdds-background-image-scania-symbol-png);
     background-size: 30px auto;
     width: 64px;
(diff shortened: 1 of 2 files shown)


theJohnnyMe:
There is something off with the scrollbar in the mobile menu. It seems that the account part of the menu hovers over the rest of the menu and there is no scrollbar. I did a video here to explain better, it is a comparison of currently released code and one from this branch: 




https://user-images.githubusercontent.com/17004063/190337679-ffc296fa-b1eb-4924-900b-dc26fd9a8005.mov

theJohnnyMe (changes requested):
The mobile menu needs to have a scrollbar function for the entire menu. Please check the video in the previous comment.

alexlioliopoulos (changes requested):
Great work indeed! From the design side, we only have some styling feedback, but I guess that otherwise the blocking of the scrollbar in the mobile menu is a quite important one that we need to solve.

**Important changes**

- The top part of the searchbar should be white and not blue. Ludwig changed that a week ago, but perhaps there was a mixup with the files.

image

should be like that:

image

All styling details you can find in the master Tegel UI Library, or give me a call.

- It seems that the search icon and the text besides it are not aligned:

image

- The mobile menu currently looks like this:

image

Ludwig would like it to be like that:

image


So as you can see, the main difference is that currently we have a "tooth" solution, having a white little square and a blue bar next to it. Ludwig wants it to be all white. Don't worry about the placement of the X icon, it can stay where you currently have it, it is no biggie.

I will come back in another comment with comments of less improtance / priority.

Let's discuss if you have any concerns!

alexlioliopoulos:
Smaller comments, or rather comments that can be fixed later on depending on how much work they will be needing:

- When there is a double line in the search suggestions, things get cluttered:

image

According to design there should be a 16px padding around the purple box in the above image.

- The search text should be black.

image

alexlioliopoulos:
Oh, wow! That was fast and great work Victor, thanks so much! 

Optional improvement that just came to mind: Could we autofocus the searchfield when clicking on the search icon? In simpler words, currently if a user presses the search icon on the header, they also need to click on the field in order to start typing. It would be much better if the search field would rightaway be focused so that users can start typing. 

If it is something easy, let's do it now. Otherwise it can of course wait.

theJohnnyMe:
Nice work Victor and a fast fix for the scrolling issue! ;) 

This is a tiny one. So when in the mobile menu when the search bar is active, you can see mobile menu items like the "Logout" button showing underneath it. Please check if the search menu can be expanded over all available height. It seems that this is the issue only when the screen height is smaller. 

Screenshot 2022-09-19 at 09 09 31

PR #442, opened 2022-09-14
fix: (toggle) changed disabled colors

victorscania:
**Describe pull-request**  
Changed colors of disabled state

**Solving issue** 
Fixes: AB#2204

**How to test**  
Check the component in storybook and inspect if the colors are correct


alexlioliopoulos (approved):
Looks good to me!

theJohnnyMe (changes requested):
Changes look good but they need to be applied in the different CSS file. So you can see that disabled color is defined by a variable called "--sdds-toggle-slider-disabled. That one can be found in : `sdds/theme/light/src/components/_toggle.scss`.

Please revert these changes and apply them in theming files as we will keep those variables for theming solution.

theJohnnyMe (approved):
@victorscania thanks for reverting and fixing changes in theming package, looks good to me ;)

PR #443, opened 2022-09-15
fix(checkbox): moving theming rules to component folder

theJohnnyMe:
**Describe pull-request**  
Moving theming rules to component folder

**Solving issue**  
Fixes: [AB#1738](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1738)

**How to test**  
1. Open Checkbox in Storybook of this branch
2. Open Checkbox in Storybook of production
3. Compare colors - there should not be any difference between those two


components/src/components/checkbox/checkbox.scss
```diff
@@ -1,16 +1,25 @@
-@import "../../helpers/style-helpers.scss";
 @import "@scania/typography/dist/scss/mixins";
 @import "@scania/typography/dist/scss/tokens";
 
-.#{$prefix}-checkbox-item {
+:root,
```
theJohnnyMe, line 4: Question for all of us: Maybe we should limit this only to the component level and not root or HTML  as it is only used in this specific component?

alexlioliopoulos:
https://user-images.githubusercontent.com/76053160/190346083-c44f68e1-2e40-4c69-b7fa-97ce0f25d1e2.mp4

To me there seems to be a slight push of the labels towards the right. But since we clearly have colour changes between the two branches (your branch seems to be building on Victor's branch where he was fixing colour issues), I am not sure if this push is because of your changes or because of Victor's changes. Let me check a bit further.

alexlioliopoulos:
**Update**

Ok it seems that Victor's branch was already merged. I think that what is actually happening is not beause of your changes but because of how Victor built the clickable aura around the component, which probably pushed the margin towards the right. I will report a separate ticket for that, but then at least we are good to go with this change here! Nice!

EmelieLitwin (approved):

PR #444, opened 2022-09-19
fix(radiobutton): moving theming rules to component folder

EmelieLitwin:
**Describe pull-request**
Moving theming rules to component folder

**Solving issue**  
Fixes: [AB#1739](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1739)

**How to test**  

1. Open radiobutton in Storybook of this branch
2. Open radiobutton in Storybook of production
3. Compare styling- there should not be any difference between those two


alexlioliopoulos:
Looks good to me! 

There is an issue with the placement/size of some of the elements, but this might be because of the "aura" task that @victorscania  worked on.

image

theJohnnyMe:
> Looks good to me!
> 
> There is an issue with the placement/size of some of the elements, but this might be because of the "aura" task that @victorscania worked on.
> 
> image
> 
> Check the 3rd circle.

@alexlioliopoulos - Something is off when the button is in a disabled state, then the position goes off a bit. Please report it in another ticket so we do not forget it.

Since it is not caused by @EmelieLitwin work, we can merge this one ;)

theJohnnyMe (approved):
Nice work Emelie ;)

PR #449, opened 2022-09-20
fix(form elements): Opened shadow DOM to be able to read the proper values

albinscania:
**Describe pull-request**  
Shadow DOM has been opened for textfield, textarea and slider.
This means that those elements can now be read as they were "proper" form elements.
A new story

**Solving issue**  
Fixes: #1896

**How to test**  
A new entry "FormTest" under "Utilities" was created, based on the original test-code from Mena.
Open the console and see what happens when the submit-buttons are pressed.


alexlioliopoulos:
Not related too much with the review you are seeking here, but here is what I get in StoryBook:

image

albinscania:
@alexlioliopoulos hmm ok that's not good! I'll have a look.

albinscania:
@alexlioliopoulos alright, i've pushed an update (give it a minute) that fixed that. 
It had nothing to do with my changes, so it's weird (but still, fixed!)

alexlioliopoulos:
Looks good, thanks!

albinscania:
@alexlioliopoulos @theJohnnyMe @EmelieLitwin 
so, my latest push actually also fixes the "disabled" textfield bug (can no longer be tabbed to).
I don't really think it's necessary to make yet another PR?

alexlioliopoulos:
> @alexlioliopoulos @theJohnnyMe @EmelieLitwin sorry my bad, it was supposed to be read-only state! (but there was also a bug with disabled, which is fixed now ;) )

Super, works great now!

Out of curiosity, what was the issue with the disabled state?

albinscania:
Also, read-only state for slider is added.
All in the same PR (will make it tons easier to merge later! sorry for any inconvienence)

alexlioliopoulos:
@matyldamc I'd like your opinion on the read-only slider.

Currently it looks exactly as the normal slider, it is just not possible to interact with it.

Since we have no design for it, do you have an opinion on how it should look like? For instance, perhaps a grey slider instead of blue?

matyldamc:
> @matyldamc I'd like your opinion on the read-only slider.
> 
> Currently it looks exactly as the normal slider, it is just not possible to interact with it.
> 
> Since we have no design for it, do you have an opinion on how it should look like? For instance, perhaps a grey slider instead of blue?

Yes, my first thought was grey (with consideration for AA), and inability to interact with the slider (no hover effects, etc)

albinscania:
@alexlioliopoulos i similiar thing with the textfield, that you could "tab into" the input field.

@matyldamc will there be design for the read-only state?

alexlioliopoulos:
@albinscania after talking to Matylda, 

> @alexlioliopoulos i similiar thing with the textfield, that you could "tab into" the input field.
> 
> @matyldamc will there be design for the read-only state?

Matylda and I discussed and for the read-only state we would like to have it as you do, so no color change, but have it un-hoverable. So no interaction can happen with it! Does it make sense?

components/src/components/textarea/textarea.tsx
```diff
@@ -1,7 +1,7 @@
 import {
   Component,
   h,
-  Listen,
+  /*Listen,*/
```
theJohnnyMe, line 4: I guess you can remove this one.

components/src/components/textfield/textfield.tsx
```diff
@@ -3,15 +3,15 @@ import {
   h,
   State,
   Prop,
-  Listen,
+  /* Listen ,*/
```
theJohnnyMe, line 6: Delete it if not needed.

theJohnnyMe:
Other than the code that is commented out, it looks good to me.

theJohnnyMe (approved):

alexlioliopoulos:
> @albinscania after talking to Matylda,
> 
> > @alexlioliopoulos i similiar thing with the textfield, that you could "tab into" the input field.
> > @matyldamc will there be design for the read-only state?
> 
> Matylda and I discussed and for the read-only state we would like to have it as you do, so no color change, but have it un-hoverable. So no interaction can happen with it! Does it make sense?

@albinscania just a reminder

PR #454, opened 2022-09-21
fix(modal): testing scrollable modal

victorscania:
**Describe pull-request**  
Suggested fix for modal scrollbar. Not fina, just for testing purpouses.

**Solving issue**  
Fixes: [AB#2226](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2226)

**How to test**  
Check the storybook link, go to modal and scroll/inspect


components/src/components/modal/_modal-core.scss
```diff
@@ -94,6 +101,27 @@ $modals: (
   position: relative;
   border-radius: 1rem;
   padding: 16px;
+
+  /* SCROLL STUFF */
+
+  /* width */
+  &::-webkit-scrollbar {
```
theJohnnyMe, line 108: See if instead of this you can use mixin from here instead: components/src/helpers/components-shared.scss

theJohnnyMe:
@victorscania thanks for making this PR quickly.
So my question for @alexlioliopoulos is and I think that we would like rather scroll only the body part of the component so the header and buttons are excluded from it.
Also, I would suggest we define the max height of the component, which could be sometime 60% or 60vh after which we get a scroll bar. What do you think guys?

alexlioliopoulos (changes requested):
Good job! I know that the task was not very clear to start with, apologies for that. Matylda and I talked about this and we would like to check if some changes could be made:

- Not sure exactly how you built it, but it seems that we go into scroll mode quite "early". Perhaps we could enter scroll mode when 80% of the available viewport area is occupied, as Johnny suggested sometime in the past.
- Could we have an option to fix the bottom bar (the one containing the buttons), the same way as we are fixing the header bar? So that the header and the buttons are always visible. But also the option to not have the bottom bar visible, for the cases when we don't have buttons.

alexlioliopoulos:
Almost identical comments Johnny!

alexlioliopoulos:
Works great! The only tiny thingy that you need to fix from my side is this:

image

on the bottom of the button bar, we can still see the content that lives behind. I guess some container needs to be pushed 1px down or something.

victorscania:
> Works great! The only tiny thingy that you need to fix from my side is this:
> 
> image
> 
> on the bottom of the button bar, we can still see the content that lives behind. I guess some container needs to be pushed 1px down or something.

Seems like a rounding error when the height of the modal lands on a sub-pixel. I've added an extra pixel to the element to avoid it. Works on my screen, but please test it

alexlioliopoulos (approved):

theJohnnyMe (approved):
Nice work Victor! Looks good to me!

PR #464, opened 2022-09-26
fix: style in radiobtn and accordion after move

EmelieLitwin:
**Describe pull-request**  
Removed use of get-colour function in component styling


alexlioliopoulos:
It's just hard to test this here, since it was always working without problems in StoryBook but the problem was in our website. Could we try how it works there?

theJohnnyMe (approved):
Thanks for sorting this one out ;)

PR #478, opened 2022-10-04
fix(header): added box-shadow and border-radius (bottom left and righ…

mJarsater:
fix(header): added box-shadow and border-radius (bottom left and right) to the dropdown-menu

Added box-shadow and border-radius (bottom left and right) to the dropdown-menu of the header component. The box-shadow was the original issue but I also noticed that the dropdown was missing a border-radius so added that fix to this commit.

**Describe pull-request**  
This PR adds the correct styling to the dropdown-menu in the header component. 

**Solving issue**  
Fixes: AB#2270

**How to test**  
1. Go to the storybook page below. 
2. Check in Header -> Inline Menu, Toolbar Menu, Searchbar Menu, All Menus and compare the styling of the dropdown to that in Figma.

**Screenshots**  
Screenshot 2022-10-04 at 11 05 04

### DIFF
--- a/components/src/components/header/header.scss
+++ b/components/src/components/header/header.scss
@@ -290,6 +290,8 @@
     transform: scaleY(0);
     transform-origin: top;
     transition: transform 150ms ease;
+    box-shadow: var(--sdds-card-box);
+    border-radius: 0 0 4px 4px;
   }
 
   .sdds-nav__dropdown-item {


theJohnnyMe:
@charlottelorensson or @matyldamc  - Please check design changes and approve if looks OK :)

components/src/components/header/header.scss
```diff
@@ -290,6 +290,8 @@
     transform: scaleY(0);
     transform-origin: top;
     transition: transform 150ms ease;
+    box-shadow: var(--sdds-card-box);
```
theJohnnyMe, line 293: Please create a local header variable for this one instead of using card one. 
I am afraid if we change anything in the card component shadow then it will propagate here too.
Before we had situations where the component was depending on the CSS of another component - mess to refactor after some time. 
Rule of thumb - component needs to be dependent on colors, typography and spacing from foundation files only.

Another, longer approach is to recognize if this type of shadow is common for many components so that we can make a global variable for it. If you decide to go this route, let's include designers so they can help to recognize in which places this kind of shadow is used. Then it could be part of shared-components.scss file.

Whatever option you decide to take is good one.

mJarsater: Sure thing, good input!

I will update the PR with a local variable for now, so that we can merge it and fix the actual bug. Then I'll do some investigation on how widely used these box-shadow stylings are.

theJohnnyMe (approved):
Look good to me. I tried to reproduce the issue Matylda was facing but failed, it looks fixed now. Maybe it was a glitch between PR builds.

mJarsater:
> Look good to me. I tried to reproduce the issue Matylda was facing but failed, it looks fixed now. Maybe it was a glitch between PR builds.

Nice, the issue was that the actual <a> seem to have some styling attached to it which also needed to border-radius specified.

matyldamc (approved):

PR #479, opened 2022-10-04
feat(modal): add prop to to hide close button

mJarsater:
Added a prop to be able to chose if to modal has a close button or not.

**Describe pull-request**  
The PR adds a prop ("hideClose") to the modal which if true hides the close button in the top right corner. 

**Solving issue**  
Fixes: [AB#2298](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2298)

**How to test**  
1. Go to the storybook page below.
2. Check in Modal and open modal number 3. 
3. See that this modal does not have a close button due to it being passed the attribute "hide-close=true".

**Screenshots**  
Screenshot 2022-10-04 at 16 16 14

### DIFF
--- a/components/src/components/modal/modal.stories.js
+++ b/components/src/components/modal/modal.stories.js
@@ -8,6 +8,13 @@ export default {
       },
       defaultValue: false,
     },
+    hideClose: {
+      name: 'Hide close button',
+      control: {
+        type: 'boolean',
+      },
+      defaultValue: false,
+    },
     size: {
       name: 'Size',
       control: {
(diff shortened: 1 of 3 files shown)


matyldamc:
@EmelieLitwin you have a good point from the user perspective. Not having the X forces the user to select one of the choices, "cancel" in this case doing the work of the X. If accessibility best practices mean always having the X, then I suggest we remove the "hide close button" functionality. Thanks for highlighting this.

theJohnnyMe:
@mJarsater - We will close this PR then as what the user asked for is not per our guidelines currently. Sorry that you spent time on this one, we would try to be better at reviewing this kind of request at an earlier stage.

mJarsater:
> @mJarsater - We will close this PR then as what the user asked for is not per our guidelines currently. Sorry that you spent time on this one, we would try to be better at reviewing this kind of request at an earlier stage.

No problemo!

Issue #362, opened 2022-04-26
[Bug report]: Mobile Nav Menu Dropdown Nav Link Active Incorrect Position


furkick:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have made sure that the issue comes from SDDS packages by removing SDDS from my project as a test.
- [X] I have followed the installation guide.

### Package

@scania/components

### Package versions

@scania/components: 4.0.0

### Browser

Chrome

### Framework

Angular

### Version

Angular 13.3.0

### Reproduction steps

1. Create a navbar with dropdown links
2. Trigger the mobile menu
3. Set a dropdown link to active
4. Observe the active indicator position 

### Screenshots

Nav Dropdown
![image](https://user-images.githubusercontent.com/7430414/165267736-aac82e8c-6b31-475a-b3f2-7f947d9a6aa4.png)

Nav Item

![image](https://user-images.githubusercontent.com/7430414/165267933-996c04a3-e3d9-419a-8250-3f5e9b2d4765.png)

### Expected behaviour

The active link indicator should be aligned with the dropdown link.

### Contact information

christopher.halfpenny@scania.com

alexlioliopoulos:
Thanks for reporting! We will soon start working on a fix for this.

alexlioliopoulos:
Hi @furkick! Does the mobile menu coming from the header, or from the side-menu component? Meaning, in the desktop version, do you or do you not have a side menu?

alexlioliopoulos:
Thanks a lot!

PR #480, opened 2022-10-04
fix(header): fix position of active item indicator when its on an inline menu

alexandereneroth:
**Describe pull-request**  
On the mobile menu, the active item indicator was displaying in the wrong position if the active item was a inline menu. This fixes that

**Solving issue**  
Fixes: #362 [AB#1895](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1895)

**How to test**  
1. Run SDDS
2. Edit Header story -> inline menu, so the item with the inline menu is the active one
3. Look at the story in Storybook

**Screenshots**  
Before:
Screenshot 2022-10-04 at 16 56 39

After:
Screenshot 2022-10-04 at 17 31 04

### DIFF
--- a/components/src/components/header/header.scss
+++ b/components/src/components/header/header.scss
@@ -169,6 +169,7 @@ html,
   }
 
   .sdds-nav__item {
+    position: relative;
     border-right: 1px solid var(--sdds-blue-700);
     border-top: none;
     border-bottom: none;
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
Looks good to me!

PR #482, opened 2022-10-05
fix(button): corrected the styling of seconday,ghost and danger button

mJarsater:
The focus style was wrong for all of the buttons above. The border should be 2px according to design, however it was 1px, needed to do some calculation to retain the correct height of the buttons on focus state. The secondary button also had the wrong background, border-color and color for hover-state.

**Describe pull-request**  
The focus style was wrong for all of the buttons above. The border should be 2px according to design, however it was 1px, needed to do some calculation to retain the correct height of the buttons on focus state. The secondary button also had the wrong background, border-color and color for hover-state.

**Solving issue**  
Fixes: AB#1820

**How to test**  
1. Go to storybook page below.
2. Check in the button component.
3. Compare the hover state of the secondary button and the focus state of secondary, ghost and danger buttons to the design (https://www.figma.com/file/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=1574%3A72148). Also test it with the different sizes of the buttons.

**Screenshots**  

**Additional context**  
Secondary hover state:
Screenshot 2022-10-05 at 10 51 13

Secondary, ghost and danger focus state:
Screenshot 2022-10-05 at 10 52 28

Screenshot 2022-10-05 at 10 52 10

Screenshot 2022-10-05 at 10 52 49


components/src/components/button/button.scss
```diff
@@ -33,7 +33,9 @@ $props: (background, border-color, color);
   }
 
   &:focus {
-    outline: none;
+    border: 2px solid;
```
theJohnnyMe, line 36: Please check if instead of trick with height and padding you can use something like the last rule from this file:
`components/src/helpers/components-shared.scss`, if focus colors are the same, then you can even use mixin here.

mJarsater: I have updated the PR to use the mixin instead. Had to add a new variable for outline-color tho, so please check that everything looks OK. 😄

Issue #347, opened 2022-04-05
[Bug report]: Avatar dropdown displays on page load, causing a flashing animation.


furkick:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have made sure that the issue comes from SDDS packages by removing SDDS from my project as a test.
- [X] I have followed the installation guide.

### Package

@scania/components

### Package versions

@scania/components: 3.0.0

### Browser

Chrome

### Framework

Angular

### Version

Angular 13.3.0

### Reproduction steps

1. Download and run the example project
2. Refresh the page
3. Observe the Avatar dropdown

### Code example

[sdds-repro.zip](https://github.com/scania-digital-design-system/sdds/files/8415728/sdds-repro.zip)

### Screenshots

![screen-capture](https://user-images.githubusercontent.com/7430414/161698422-c4173cc2-ea42-4fb7-ad8d-48d957afcfa8.gif)

### Expected behaviour

The Avatar dropdown should remain hidden until it has been actioned to display.

### Contact information

christopher.halfpenny@scania.com

alexlioliopoulos:
Hi @furkick ! It might be the case that our demo page does not come with the latest header component. We are currently working on a new demo page.

Meanwhile, have a look at our header component in Storybook, in case it helps somehow.

https://production.d1g8v8mrkx992n.amplifyapp.com/?path=/story/patterns-header--all-menus

furkick:
Hi @alexlioliopoulos, I have had a look at the Storybook and it seems to suffer from the same issue when loading?

![screen-capture](https://user-images.githubusercontent.com/7430414/163393091-448958ad-e8a3-4028-8dc3-2aa0a55604a6.gif)

The example above shows that multiple components flash/load on page load as opposed to being hidden until actioned.

alexlioliopoulos:
Aha, yes you are right! Let us have a look at it. Thanks for reporting it btw.

PR #484, opened 2022-10-05
fix(header): stop dropdowns showing briefly on page load

alexandereneroth:
**Describe pull-request**  
Will hide the header menus with `opacity: 0` when they are not open, so they wont show briefly when the page loads. 
See: #347

This is a partial quick fix to a larger issue. 
The issue seems to happen on anything with a transition property defined on it. If i add a background color transition on sdds-btn, for example, it will fade in from the browser defaults on page load. 
I think its because the css is loaded in sdds-theme, so it might come in a bit late.

re #347, [AB#1931](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/1931)

**How to test**  
Go to this page, and refresh the page: https://production.d1g8v8mrkx992n.amplifyapp.com/iframe.html?args=&id=component-header--all-menus&viewMode=story

### DIFF
--- a/components/src/components/header/header.scss
+++ b/components/src/components/header/header.scss
@@ -290,12 +290,14 @@ html,
 
     .sdds-nav__dropdown-menu {
       transform: scaleY(1);
+      opacity: 1;
     }
   }
 
   .sdds-nav__dropdown-menu {
     position: absolute;
     list-style-type: none;
+    opacity: 0;
     transform: scaleY(0);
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
Nice trick!

albinscania (approved):
Makes sense!

PR #485, opened 2022-10-05
fix(dropdown): place outside label 8px above dropdown like design spec

alexandereneroth:
**Describe pull-request**  
Places outside label variant of dropdown 8px above the dropdown, to align with appearance in figma, and also with appearance of text field. 
Also makes the label take up space on the page, which it did not previously do (it was placed with position absolute).

**Solving issue**  
[AB#718](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/718)

**How to test**  
Go here https://production.d1g8v8mrkx992n.amplifyapp.com/?path=/story/component-dropdown--label-outside,
or equivalent path on your localhost when running sdds.

**Screenshots**  
### Before
Screenshot 2022-10-05 at 16 19 48

### After
Screenshot 2022-10-05 at 16 19 56

### DIFF
--- a/components/src/components/dropdown/_dropdown-core.scss
+++ b/components/src/components/dropdown/_dropdown-core.scss
@@ -78,7 +78,7 @@ html {
 .sdds-dropdown-label-outside {
   @include type-style("detail-05");
 
+  display: block;
   color: var(--sdds-grey-958);
-  position: absolute;
-  top: -16px;
+  margin-bottom: var(--sdds-spacing-element-8);
 }
--- a/components/src/components/dropdown/dropdown.scss
+++ b/components/src/components/dropdown/dropdown.scss
@@ -121,6 +121,12 @@
     }
   }
 
+  &:has(span.sdds-dropdown-label-outside) {
+    .sdds-dropdown-menu {
+      margin-bottom: -24px;
+    }
+  }
+
   .sdds-dropdown-menu {
     z-index: 10000;
 
--- a/components/src/components/dropdown/dropdown.scss
+++ b/components/src/components/dropdown/dropdown.scss
@@ -121,12 +121,6 @@
     }
   }
 
-  &:has(span.sdds-dropdown-label-outside) {
-    .sdds-dropdown-menu {
-      margin-bottom: -24px;
-    }
-  }
-
   .sdds-dropdown-menu {
     z-index: 10000;
 
@@ -181,6 +175,12 @@
       box-shadow: 0 -1px 3px 0 rgb(0 0 0 / 10%);
       transform-origin: bottom;
     }
+  }
+}
+
+:host(.sdds-dropdown--open-upwards[label-position="outside"]) {
+  span.sdds-dropdown-menu {
+    bottom: calc(100% - 24px);
   }
 }
 


theJohnnyMe:
If there is a lack of space at the bottom, the dropdown menu will open upwards.
It causes this behavior:
Screenshot 2022-10-06 at 09 22 57

@charlottelorensson and @matyldamc  - Should the menu in this case touch the top part of the dropdown and hide the label?

matyldamc:
@theJohnnyMe unfortunately yes, it should behave as you said, and cover the label. Not great for accessibility but I see no other solution.

albinscania (approved):

theJohnnyMe (dismissed):
Please check Matylda's comment.

mJarsater:
@theJohnnyMe I did a small update to remove the gap, please have a look whenever you have time 😄

theJohnnyMe:
I am just thinking if overwriting the existing rule for a situation where the label is out would be a more neat solution, so instead of introducing -24px on the margin, can you change the bottom property calculation in case the label is out? 

I did test it here, seems to work OK:

Screenshot 2022-10-10 at 11 52 53

so something like this in the code: 

`
:host(.sdds-dropdown--open-upwards:has(span.sdds-dropdown-label-outside)) {
  .sdds-dropdown {
    .sdds-dropdown-menu {   
      bottom: calc(100% - 24px);     
    }
  }
}
`

What do you think? 🤔

mJarsater:
> I am just thinking if overwriting the existing rule for a situation where the label is out would be a more neat solution, so instead of introducing -24px on the margin, can you change the bottom property calculation in case the label is out?
> 
> I did test it here, seems to work OK:
> 
> Screenshot 2022-10-10 at 11 52 53
> 
> so something like this in the code:
> 
> `:host(.sdds-dropdown--open-upwards:has(span.sdds-dropdown-label-outside)) { .sdds-dropdown { .sdds-dropdown-menu { bottom: calc(100% - 24px); } } }`
> 
> What do you think? 🤔

I definitely like this approach more than the negative margin!

theJohnnyMe:
> > I am just thinking if overwriting the existing rule for a situation where the label is out would be a more neat solution, so instead of introducing -24px on the margin, can you change the bottom property calculation in case the label is out?
> > I did test it here, seems to work OK:
> > Screenshot 2022-10-10 at 11 52 53
> > so something like this in the code:
> > `:host(.sdds-dropdown--open-upwards:has(span.sdds-dropdown-label-outside)) { .sdds-dropdown { .sdds-dropdown-menu { bottom: calc(100% - 24px); } } }`
> > What do you think? 🤔
> 
> I definitely like this approach more than the negative margin! Me and @alexandereneroth was discussing the :has selector tho, it is not yet supported in FF so we would have to look use something else to check if the label is outside and then target the parent element.

I get you, good thinking for support on Firefox. Well, since it is a web component, one way is that if the label is used and it is lable outside, you can make another class to appear on top level, something like "sdds-dropdown--lable-outside" and based on that style menu position accordingly, just one idea.

mJarsater:
> > > I am just thinking if overwriting the existing rule for a situation where the label is out would be a more neat solution, so instead of introducing -24px on the margin, can you change the bottom property calculation in case the label is out?
> > > I did test it here, seems to work OK:
> > > Screenshot 2022-10-10 at 11 52 53
> > > so something like this in the code:
> > > `:host(.sdds-dropdown--open-upwards:has(span.sdds-dropdown-label-outside)) { .sdds-dropdown { .sdds-dropdown-menu { bottom: calc(100% - 24px); } } }`
> > > What do you think? 🤔
> > 
> > 
> > I definitely like this approach more than the negative margin! Me and @alexandereneroth was discussing the :has selector tho, it is not yet supported in FF so we would have to look use something else to check if the label is outside and then target the parent element.
> 
> I get you, good thinking for support on Firefox. Well, since it is a web component, one way is that if the label is used and it is lable outside, you can make another class to appear on top level, something like "sdds-dropdown--lable-outside" and based on that style menu position accordingly, just one idea.

I think we can target the right element like this: 

```
:host(.sdds-dropdown--open-upwards[label-position="outside"]){
   span.sdds-dropdown-menu { 
      bottom: calc(100% - 24px); 
    } 
}
```

@alexandereneroth do you want to test it out aswell?

theJohnnyMe:
> > > > I am just thinking if overwriting the existing rule for a situation where the label is out would be a more neat solution, so instead of introducing -24px on the margin, can you change the bottom property calculation in case the label is out?
> > > > I did test it here, seems to work OK:
> > > > Screenshot 2022-10-10 at 11 52 53
> > > > so something like this in the code:
> > > > `:host(.sdds-dropdown--open-upwards:has(span.sdds-dropdown-label-outside)) { .sdds-dropdown { .sdds-dropdown-menu { bottom: calc(100% - 24px); } } }`
> > > > What do you think? 🤔
> > > 
> > > 
> > > I definitely like this approach more than the negative margin! Me and @alexandereneroth was discussing the :has selector tho, it is not yet supported in FF so we would have to look use something else to check if the label is outside and then target the parent element.
> > 
> > 
> > I get you, good thinking for support on Firefox. Well, since it is a web component, one way is that if the label is used and it is lable outside, you can make another class to appear on top level, something like "sdds-dropdown--lable-outside" and based on that style menu position accordingly, just one idea.
> 
> I think we can target the right element like this:
> 
> ```
> :host(.sdds-dropdown--open-upwards[label-position="outside"]){
>    span.sdds-dropdown-menu { 
>       bottom: calc(100% - 24px); 
>     } 
> }
> ```
> 
> @alexandereneroth do you want to test it out aswell?

Great idea!

alexandereneroth:
@mJarsater That worked well, nice!

alexandereneroth:
@theJohnnyMe Should be fixed now

mJarsater (approved):
Looks good to me!

theJohnnyMe (approved):
Nice ideas and great work!

PR #486, opened 2022-10-07
fix(tabs): Calculate tabwidth (and auto height) in componentDidRender()-method

albinscania:
Description says it all.

**How to test**  
Open the react project in the sandbox. Reload the page and make sure the tabs are of equal width.

### DIFF
--- a/components/src/components/tabs/inline-tabs-fullbleed/inline-tabs-fullbleed.tsx
+++ b/components/src/components/tabs/inline-tabs-fullbleed/inline-tabs-fullbleed.tsx
@@ -24,9 +24,12 @@ export class InlineTabsFullbleed {
 
   buttonWidth: number = 0; // current calculated width of the largest nav button
 
+  componentDidRender() {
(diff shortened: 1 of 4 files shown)


albinscania:
Oh! Don't forget to LINK first, to actually try it out in the react sandbox.

mJarsater:
Might I suggest that we add a cursor:no-drop to the :hover on the disabled tabs, to correspond with the style of the disabled buttons. But we might need to check that with design first?

albinscania:
> 

that's a great idea! will fix that on monday.

theJohnnyMe:
@AlbinFarstun  - I think the name of PR is misleading, I guess it is about "inline-tabs".

mJarsater:
Got the linking to work now and it looks great!

Would be nice if we could add the cursor:not-allowed on the disabled tabs and have the style not be inline. But that could be a future PR. 
Screenshot 2022-10-10 at 16 01 31

theJohnnyMe (approved):

PR #488, opened 2022-10-10
chore: reduce noise when running npm start

alexandereneroth:
**Describe pull-request**  
Adds --quiet flag to `storybook-start` command to reduce the amount of unnecessary noise that appears in the console when running `npm start`

**How to test**  
`npm start`

**Additional context**  
There was previously too much output in the console, to the point where it was a real chore to look through it.


theJohnnyMe (approved):
Looks good to me

PR #495, opened 2022-10-12
feat(link): migrate link component

alexandereneroth:
Migrate link component

Fixes: [AB#2399](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2399)

### DIFF
--- a/components/src/components/link/link.stories.ts
+++ b/components/src/components/link/link.stories.ts
@@ -1,0 +1,55 @@
+import { Meta, StoryFn } from '@storybook/html';
+import format from 'html-format';
+import clsx from 'clsx';
+
+const defaultArgs = {
+  disabled: false,
+  noUnderline: false,
+};
+
+type Args = typeof defaultArgs;
+
+export default {
+  title: 'Components/Link',
+  argTypes: {
+    disabled: {
+      name: 'Disabled',
+      description: 'Display link in disabled state',
+      defaultValue: false,
+      type: 'boolean',
+    },
+    noUnderline: {
+      name: 'No underline',
+      description: 'Hide underline under link text',
+      defaultValue: false,
+      type: 'boolean',
+    },
+  },
+} as Meta<Args>;
+
+const Template: StoryFn<Args> = args => {
+  return format(
+    `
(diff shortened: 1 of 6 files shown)


mJarsater (approved):
Looks good to me

components/src/components/link/link.stories.ts
```diff
@@ -0,0 +1,55 @@
...
+  ...Default.args,
+  disabled: true,
+};
+
+export const NoUnderline = Template.bind({});
+
+NoUnderline.args = {
+  ...Default.args,
+  noUnderline: true,
+};
```
mJarsater, line 55: Should this be a new story? Or would we want to streamline it and let the users view this via the controls.

alexandereneroth: I feel like there's value to displaying the different variations to the user up front.
Some examples:
- https://digitaldesign.scania.com/components/datepicker/code 
- https://chakra-ui.com/docs/components/link/usage
- https://mui.com/joy-ui/react-link/

mJarsater: Absolutely, I think this also is a question of how user actually use the site. Are they interacting with the controls or are they checking the stories.

theJohnnyMe (changes requested):
Please leave only the "Default" story in this one. It is a very simple component. Thanks for fixing the disabled state too.

alexandereneroth:
> Please leave only the "Default" story in this one. It is a very simple component. Thanks for fixing the disabled state too.

Removed the other stories. 
Also added a demo class with a font size so the text is not tiny.

components/src/components/link/link.stories.ts
```diff
@@ -0,0 +1,48 @@
+import { Meta, StoryFn } from '@storybook/html';
+import clsx from 'clsx';
+import { formatHtmlPreview } from '../../utils/utils';
+
+const defaultArgs = {
```
mJarsater, line 5: Is this object needed? The default values are already present in the default export.

PR #497, opened 2022-10-12
docs: make docs tab default in storybook

alexandereneroth:
Not sure what was decided, but if we decided to change the default tab we can merge this PR.

**Describe pull-request**  
Makes the docs tab the default tab in storybook, instead of canvas.

**Screenshots**  
Screenshot 2022-10-12 at 12 07 45

### DIFF
--- a/components/.storybook/preview.js
+++ b/components/.storybook/preview.js
@@ -1,3 +1,4 @@
+import { themes } from '@storybook/theming';
 import { defineCustomElements } from '../loader';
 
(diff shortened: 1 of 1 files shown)


mJarsater:
I would say that this is dependant on what audience our storybook has. Is it designer or developers?

mJarsater (approved):

theJohnnyMe (approved):
Tested, looks good to me!

PR #498, opened 2022-10-12
docs: show code example by default in storybook

alexandereneroth:
Shows code example by default in storybook.
Thought it felt better..
What do you think?

**Screenshots**  
Screenshot 2022-10-12 at 12 18 14

### DIFF
--- a/components/.storybook/preview.js
+++ b/components/.storybook/preview.js
@@ -47,6 +47,11 @@ export const parameters = {
     lightClass: 'sdds-theme-light',
     stylePreview: true,
   },
+  docs: {
+    source: {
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
Looks good to me

alexandereneroth:
> Looks good to me, but we may need to look out for some components with a lot of code/controls and several stories so that the users don't have to scroll forever 😆

Yes I agree! 
I think in most component library documentations they show code, but then have a "show more..." button or simiar, which i guess is a middle ground between showing all code and no code. 
Not sure that is possible in storybook without markdown documentation though 😕

PR #503, opened 2022-10-13
feat(spinner): migrate spinner

EmelieLitwin:
**Describe pull-request**  
Migrating spinner component


components/src/global/global.scss
```diff
@@ -25,6 +25,7 @@
 
 //CSS component
 @import "src/components/banner/banner";
+@import "src/components/spinner/spinner";
```
mJarsater, line 28: Since it's a pure webcomponent I don't think this needs to be imported globally?

PR #505, opened 2022-10-13
feat(datetime): migrate datetime component

alexandereneroth:
Migrate datetime

Fixes: [AB#2400](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2400)

**How to test**  
Check datetime in storybook


components/src/components/datetime/datetime-theme-vars.scss
```diff
@@ -0,0 +1,83 @@
+//TODO reuse textfield variables instead
```
mJarsater, line 1: Should we do this? Or should we try and encapsulate all styles to a component.

mJarsater (approved):

components/src/components/datetime/datetime.stories.ts
```diff
@@ -0,0 +1,143 @@
...
+        type: 'radio',
+        options: ['datetime-local', 'date', 'time'],
+      },
+    },
+    size: {
+      name: 'Size',
+      description: 'Switch between different sizes',
+      control: {
+        type: 'radio',
+        options: ['Large', 'Medium', 'Small'],
```
mJarsater, line 23: I would say that 'md' and 'sm' makes more sense as options since thats the prop we are passing to the component.

albinscania (approved):

components/src/components/datetime/datetime.stories.ts
alexandereneroth, line 23:
```suggestion
// todo: make consistent with other sizes, for example 'xs', 'sm', etc
        options: ['Large', 'Medium', 'Small'],
```

PR #509, opened 2022-10-14
feat(textarea/textfield): migrate textarea and textfield components

mJarsater:
**Describe pull-request**  
Migration of the textfield and the textarea components.

**Solving issue**  
Fixes: [AB#2409](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2409)

**How to test**  
1. Go to storybook link below
2. Check in components -> textarea/textfield
3. Check that the design is aligned with the one in Figma. 

**Screenshots**  
-

**Additional context**  
-


albinscania (commented):
Textarea:

"Disabled" shows a blue line when clicked on, we might want the same state here as "disabled" ?

"Label inside" gets two labels on top of each other when the textarea is selected. (might be a story tweak only maybe?)

mJarsater:
> Textfield:
> 
> "Icon" story doesn't get an icon, but this is probably because new icons are not merged yet?

Yeah, exactly.

mJarsater:
> Textarea:
> 
> "Disabled" shows a blue line when clicked on, we might want the same state here as "disabled" ?
> 
> "Label inside" gets two labels on top of each other when the textarea is selected. (might be a story tweak only maybe?)

nice catch, it's the padding not being applied correctly. I'll fix it 👍🏻

mJarsater:
Should be fixed now.

PR #514, opened 2022-10-18
feat(accordion): migrate accordion component

alexandereneroth:
Migrate accordion component

Fixes: [AB#2383](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2383)

### DIFF
--- a/components/src/components/accordion/accordion.scss
+++ b/components/src/components/accordion/accordion.scss
@@ -1,0 +1,238 @@
+:host {
+  .sdds-on-white-bg {
+    --sdds-accordion-bg-hover: var(--sdds-accordion-bg-hover-on-white);
+  }
+}
+
+// TODO question: should we not share this like before?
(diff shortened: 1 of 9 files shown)


components/src/components/accordion/accordion.scss
```diff
@@ -0,0 +1,238 @@
+:host {
+  .sdds-on-white-bg {
+    --sdds-accordion-bg-hover: var(--sdds-accordion-bg-hover-on-white);
+  }
+}
+
+// TODO question: should we not share this like before?
+@mixin sdds-focus-state {
```
mJarsater, line 8: I agree that we might wanna move this to some sort of global stylesheet that we can pull from.

mJarsater (approved):
Looks good to me. Maybe bring up the question regarding focus-state mixin on devcheckin?

PR #523, opened 2022-10-21
feat(gitignore): added *.d.ts to the gitignore

mJarsater:
**Describe pull-request**  
added *.d.ts to the gitignore

**Solving issue**  
Fixes: -

**How to test**  
-
**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/tegel/package.json
+++ b/tegel/package.json
@@ -26,7 +26,7 @@
     "generate": "stencil generate",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "deploy-storybook": "npm run install && npm run build && npm run build-storybook",
+    "deploy-storybook": "npm install && npm run build && npm run build-storybook",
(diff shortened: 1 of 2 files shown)


theJohnnyMe (approved):

tegel/package.json
```diff
@@ -26,6 +26,7 @@
     "generate": "stencil generate",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
+    "deploy-storybook": "npm install && npm run build && npm run build-storybook",
```
alexandereneroth, line 29: Is this intended to be here?
What is it for?

alexandereneroth:
I would also delete the ignored files, so a old one that is committed to the repo is not used mistakenly

mJarsater:
> I would also delete the ignored files, so a old one that is committed to the repo is not used mistakenly

I don't think the the component.d.ts has ever been committed (?), so we should be good there already.

PR #531, opened 2022-10-21
feat(button): migrate button component

mJarsater:
**Describe pull-request**  
Migration on button component

**Solving issue**  
Fixes: AB#2398

**How to test**  
1. Go to storybook link below
2. Check in components -> button
3. Check that the design is aligned with the one in Figma. 

**Screenshots**  
-

**Additional context**  
-


alexandereneroth (changes requested):

tegel/src/components/button/button.stories.js
```diff
@@ -0,0 +1,192 @@
...
+
+export const Fullbleed = ButtonTemplate.bind({});
+Fullbleed.args = {
+  text: 'Button',
+  fullbleed: true,
+  icon: true,
+};
+
+/** Button as web component */
+export const sddsButton = ComponentBtn.bind({});
```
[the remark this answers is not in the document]
mJarsater: I think this is also a bigger issue on how we display components that are both native and wc.

tegel/src/components/button/button-theme-vars.scss
```diff
@@ -0,0 +1,161 @@
...
+  --sdds-btn-danger-border-color-active: transparent;
+  --sdds-btn-danger-background-focus: var(--sdds-red-400);
+  --sdds-btn-danger-color-focus: var(--sdds-white);
+  --sdds-btn-danger-border-color-focus: var(--sdds-red-700);
+  --sdds-btn-danger-outline-color-focus: var(--sdds-red-700);
+  --sdds-btn-danger-background-disabled: var(--sdds-white);
+  --sdds-btn-danger-color-disabled: rgb(0 0 0 / 38%);
+  --sdds-btn-danger-border-color-disabled: transparent;
+
+  .sdds-on-white-bg {
```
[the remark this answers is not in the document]
mJarsater: Sure, but aren't on-white sort of a theming? That's how I interpret them at least. Let's check with the rest of the team?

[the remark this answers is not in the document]
alexandereneroth: Sure, but isn't it still a property of the button rather than the overall theme of the site?
But yea, good to bring this subject up with the team

tegel/src/components/button/button-theme-vars.scss
```diff
@@ -0,0 +1,157 @@
...
+    --sdds-btn-primary-color-focus: var(--sdds-white);
+    --sdds-btn-primary-border-color-focus: var(--sdds-blue-600);
+    --sdds-btn-primary-outline-color-focus: var(--sdds-blue-600);
+    --sdds-btn-primary-background-disabled: var(--sdds-white);
+    --sdds-btn-primary-color-disabled: var(--sdds-grey-400);
+    --sdds-btn-primary-border-color-disabled: transparent;
+
+    //Primary - On white
+    --sdds-btn-primary-on-white-background-disabled: var(--sdds-grey-50);
+    --sdds-btn-danger-on-white-background-disabled: var(--sdds-grey-50);
```
mJarsater, line 101: @alexandereneroth like this?

alexandereneroth: Yea, exactly, thats what i had in mind

tegel/src/components/button/button-theme-vars.scss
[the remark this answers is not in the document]
theJohnnyMe: I like how you give feedback to each other. Kudos!

tegel/src/components/button/button-component.tsx
```diff
@@ -0,0 +1,43 @@
...
+})
+export class SddsButton {
+  // TODO: Make this mandatory prop. Send warning to user if empty and it is not icon only version of button.
+  /** Text inside a button */
+  @Prop() text: string = '';
+
+  @Prop() type: string;
+
+  @Prop() size: 'sm' | 'md' | '' = '';
+
```
mJarsater, line 16: Should have a onwhite/ongrey prop?

theJohnnyMe:
Lets try this: 
Screenshot 2022-10-25 at 14 13 50

tegel/src/components/button/button-component.tsx
```diff
@@ -0,0 +1,51 @@
...
+  @Prop() text: string = '';
+
+  //TODO: Should this have prmary, danger, warning etc?
+  /** Type of button */
+  @Prop() type: 'primary' | 'secondary' | 'ghost' | 'danger' = 'primary';
+
+  // Button variant
+  @Prop() variant: 'default' | 'variant' = 'default';
+
+  @Prop() size: 'sm' | 'md' | 'lg' = 'lg';
```
theJohnnyMe, line 20: Please, check Figma, I think we also have "extra small" - we have in the Usage/Style tab on the website, but we are missing it here since... well since forever so it makes sense to sort it once for while.

theJohnnyMe (changes requested):
Tiny think on sizes. Let's get in touch tomorrow regarding the icon component and button - makes sense to sort it too after months of waiting. 

Please, grab me tomorrow after stand-up to let you know more about it.

tegel/src/components/button/button-component.tsx
mJarsater: Sure thing, I'll check in Figma 😄

tegel/src/components/button/button.scss
```diff
@@ -0,0 +1,218 @@
...
+    padding: var(--sdds-spacing-element-12);
+    height: $btn-sm-height;
+
+    &:focus {
+      // TODO: decide on what tod o with these @include sdds-focus-state;
+      outline: 2px solid var(--sdds-blue-400);
+      outline-offset: -2px;
+    }
+
+    span + .sdds-btn-icon {
```
mJarsater, line 81: Ideally I would like to add a class to this span to not just target the element. But that would be a breaking change. Thoughts?

tegel/src/components/button/button.scss
```diff
@@ -0,0 +1,218 @@
...
+      margin-left: var(--sdds-spacing-element-20);
+      width: var(--sdds-spacing-element-20);
+      height: var(--sdds-spacing-element-20);
+    }
+  }
+
+  .sdds-btn-lg {
+    ::slotted([slot='icon']) {
+      // TODO - This should be a var value
+      margin-left: 7rem;
```
mJarsater, line 213: This is what I found in Figma. But maybe we should ask the designer why this isn't on of our default spacing values?

tegel/src/components/button/button.scss
```diff
@@ -0,0 +1,218 @@
...
+$btn-sm-height: 10rem;
+$btn-xs-height: 8rem;
+$types: (primary, secondary, ghost, danger);
+$props: (background, border-color, color, outline-color);
+
+@mixin button-base {
+  font: var(--sdds-detail-02);
+  letter-spacing: var(--sdds-detail-02-ls);
+  display: flex;
+  align-items: center;
```
mJarsater, line 19: I found these necessary to align the icons to the center vertically. This had display inline-block before.

tegel/src/components/button/button.scss
```diff
@@ -0,0 +1,218 @@
...
+    text-align: center;
+
+    .sdds-btn-icon,
+    &.sdds-btn-icon {
+      margin-left: var(--sdds-spacing-element-8);
+    }
+  }
+
+  span {
+    align-self: baseline;
```
mJarsater, line 129: This is to adjust for the line-height of our font. This alignes it to the center vertically. This span had no display value before.

tegel/src/components/button/button.scss
theJohnnyMe: I saw there is a text class for span here, I guess you added it and I am all for targeting elements with class instead of element tag. It might be breaking change, but go for it: 

Screenshot 2022-10-28 at 09 24 00

tegel/src/components/button/button.scss
theJohnnyMe: Note to you, please check how buttons will be displayed in the modal. The button component is displayed in the Modal component and it should be in one line. Please make a ticket in "Fresh Stencil & Storybook" to double-check check Modal buttons after this merge.  I am tagging @victorscania for heads up too.

tegel/src/components/button/button.scss
theJohnnyMe: I guess here you can add a class name for text instead too or?

tegel/src/components/button/button.scss
mJarsater: Yeah, will do! Thanks 😄

tegel/src/components/button/button.scss
theJohnnyMe: Yes, I agree, please bring it up at the Dev&Design meeting. Until then, please use 28px instead as rem is not stable for us and all upper spacing variables rely on PX values too.

theJohnnyMe:
Check the icon color here, same is for "Ghost" type too:
Screenshot 2022-10-28 at 09 18 34

theJohnnyMe:
Icon on "Small" size variation needs to be 16x16px. I think you would probably have to adjust the padding in this one too to make it sure that in "Only icon" variation button is always square, I think small one is 40x40, medium 48x48 and large one 56x56. Maybe you will have to use weird values for padding that are not aligning with logic of increments of 4px, but I would say prio is to make sure icon is in the middle and button size is correct, you are free to use whatever trick to archive that :D 

Screenshot 2022-10-28 at 09 20 56

theJohnnyMe:
Related to sizing and size of the icons and padding, maybe just setting the icon 16x16 will fix the horizontal alignment between text and icon? 

Screenshot 2022-10-28 at 09 21 34

theJohnnyMe:
Same for this one: 

Screenshot 2022-10-28 at 09 21 56

theJohnnyMe:
Thanks for the great work, some issues you are fixing were there waiting for too long, so it is not only migration but finally fixing these tissues too + button exists as HTML&CSS and WebComponent which complicates things too.

mJarsater:
> Thanks for the great work, some issues you are fixing were there waiting for too long, so it is not only migration but finally fixing these tissues too + button exists as HTML&CSS and WebComponent which complicates things too.

Yeah, like the header there are a lot of changes going on at once. So it's really a good thing we have these PR's to have a review of everything. I fixed the issues that you mentioned above and I also realised we were missing some states for the icon, like the secondary + icon with hover and focus. But that should be fixed now.

theJohnnyMe:
A couple of small things found:

**Danger variation and icon color (on hover it looks good):**
Screenshot 2022-10-31 at 10 30 24

**Check the name:** 😋
Screenshot 2022-10-31 at 10 32 39

**Only icon sizing:**
The "Only icon" version needs to be square, so the large one is 56x56 etc (more in Figma file). Right now width seems to have 2-4px extra compared to the Figma file.

mJarsater:
> **Only icon sizing:**
> The "Only icon" version needs to be square, so the large one is 56x56 etc (more in Figma file). Right now width seems to have 2-4px extra compared to the Figma file.

Really nice catch! This is due to us having different padding on the icon vs the text of the button. The button for large and medium has 2px less padding on the icon vs the text. My first idea would be to solve this with a negative margin to remove the pixels that are the diff between the two. Any other thoughts?

mJarsater:
@theJohnnyMe I have added a fix now that uses negative margin to correct the size of the buttons. Can't really find any other way of doing it without redoing a bigger chunk of the component.

alexandereneroth (changes requested):

tegel/src/components/button/button-webcomponent.stories.ts
```diff
@@ -0,0 +1,135 @@
...
+    },
+  },
+};
+
+const WebComponentTemplate = ({ onlyIcon, size, variant, btnType, fullbleed, disabled, icon, text = 'Button' }) => {
+  const inlineStyle = fullbleed ? 'style="width:100%;"' : '';
+  const variantValue = variant === 'Variant' ? 'variant' : 'default';
+  return formatHtmlPreview(
+    `
+  <sdds-button onlyIcon="${onlyIcon}" type="${btnType}" size="${size}" ${disabled ? 'disabled' : ''} ${fullbleed ? 'fullbleed' : ''} text="${
```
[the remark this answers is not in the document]
mJarsater: Sure, but the prop need a name change before this can be commited.

tegel/src/components/button/button.scss
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  &.sdds-btn-fullbleed {
+    width: inherit;
+    display: flex;
+    justify-content: center;
+
+    .sdds-btn-icon,
+    &.sdds-btn-icon {
+      margin-left: var(--sdds-spacing-element-8);
```
alexandereneroth, line 143: I don't think this rule is taking effect.
However in the design fullbleed buttons have less distance between the text and icon 🤔

alexandereneroth (changes requested):

tegel/src/components/button/button.scss
mJarsater: Nice catch. I'll update the PR.

tegel/src/components/button/button.scss
```diff
@@ -0,0 +1,279 @@
...
+
+.sdds-on-white-bg {
+  --sdds-btn-primary-background-disabled: var(--sdds-btn-primary-on-white-background-disabled);
+  --sdds-btn-danger-background-disabled: var(--sdds-btn-danger-on-white-background-disabled);
+}
+
+$btn-lg-height: 14rem;
+$btn-md-height: 12rem;
+$btn-sm-height: 10rem;
+$btn-xs-height: 8rem;
```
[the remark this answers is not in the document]
mJarsater: fixed.

tegel/src/components/button/button.scss
mJarsater: fixed.

alexandereneroth (approved):
Nice work 🎉

tegel/src/components/button/button.scss
```diff
@@ -0,0 +1,318 @@
+@import './button-theme-vars.scss';
+
+.sdds-on-white-bg {
+  --sdds-btn-primary-background-disabled: var(--sdds-btn-primary-on-white-background-disabled);
+  --sdds-btn-danger-background-disabled: var(--sdds-btn-danger-on-white-background-disabled);
+}
+
+$btn-lg-height: 14rem;
```
theJohnnyMe, line 8: Please do the height in the pixels, we might edit or remove base value in future, so to be on safe spot lets use pixels instead.

theJohnnyMe (dismissed):

tegel/src/components/button/button.scss
mJarsater: Fixed, thanks!

PR #542, opened 2022-10-25
fix(textarea/textfield): handle null values textfield/textarea

mJarsater:
**Describe pull-request**  
Added null checks to allow for null values on our inputfields. With this the component will render even though it's value might be null. 

**Solving issue**  
Fixes: #526 [AB#2445](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2445)

**How to test**  
1. Checkout the branch and cd to textarea/textfield component.
2. Change de default value of the components to null
3. Check that the components render without issues.

**Screenshots** 
-

**Additional context**  
-


theJohnnyMe:
Nice trick with "?", I think we should have it in more places in our code in general.


I have one question that I am not 100% sure is right so I need another dev brain on this: 

We have prop defined like this: 
Screenshot 2022-10-31 at 10 57 16

What if we would define it like this instead: 

`
@Prop() value: string | number = null;
`

I wonder if that would be any better? Also, I do not like that in code we have prop without type, in the worst case maybe we can put "any", maybe it defaults to that if not defined, but again.

theJohnnyMe (approved):
LGTM ;)

mJarsater:
> Nice trick with "?", I think we should have it in more places in our code in general.
> 
> I have one question that I am not 100% sure is right so I need another dev brain on this:
> 
> We have prop defined like this: Screenshot 2022-10-31 at 10 57 16
> 
> What if we would define it like this instead:
> 
> `@Prop() value: string | number = null;`
> 
> I wonder if that would be any better? Also, I do not like that in code we have prop without type, in the worst case maybe we can put "any", maybe it defaults to that if not defined, but again.

Yeah this should work aswell I think. Then we wouldn't have to check for null values at all.

PR #557, opened 2022-11-01
fix(textarea/textfield): handle-null-values-textfield/textarea

mJarsater:
**Describe pull-request**  
A bug was found where to textarea/textfield components would not render in case they had null as their value. This is solved by instead adding a type to this value prop of string. This PR also updated the type-prop of the textfield component to expect 'text' or 'password'.

**Solving issue**  
Fixes: #526

**How to test**  
1. Go to storybook link below
2. Check in componets -> textarea/textfield
3. Check that the component renders even though value is initially set to null. 

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/components/src/components/textarea/textarea.tsx
+++ b/components/src/components/textarea/textarea.tsx
@@ -31,7 +31,7 @@ export class Textarea {
   @Prop() placeholder: string = '';
 
   /** Value of the input text */
-  @Prop() value = '';
+  @Prop() value: string = null;
 
   /** Set input in disabled state */
   @Prop() disabled: boolean = false;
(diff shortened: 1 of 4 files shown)


components/src/components/textarea/textarea.tsx
```diff
@@ -169,7 +169,7 @@ export class Textarea {
         )}
         {this.maxlength > 0 && (
           <div class={'sdds-textarea-textcounter'}>
-            {this.value.length}{' '}
```
theJohnnyMe, line 172: Maybe here this one is needed, it shows number of characters, right? Please double check.

components/src/components/textfield/textfield.tsx
```diff
@@ -215,7 +215,7 @@ export class Textfield {
 
           {this.maxlength > 0 && (
             <div class="sdds-textfield-textcounter">
-              {this.value.length}
```
theJohnnyMe, line 218: Maybe here this one is needed, it shows a number of characters, right? Please double-check.

theJohnnyMe (changes requested):

theJohnnyMe:
PS: PR title, I think "(button)" is misleading as this one fixes textarea/textfield, I think you can drop it and just leave "fix: ..."

components/src/components/textarea/textarea.tsx
mJarsater: Yes, of course!

alexandereneroth (approved):
Looks good to me! 🍰

PR #559, opened 2022-11-02
fix(textarea/textfield):textarea/field to handle null values for tegel

mJarsater:
This PR updates tegel with updates Tegel to solve a bug that was found in 4_10.0.

PR that has been merged in 4_10.0: https://github.com/scania-digital-design-system/sdds/pull/557

**Describe pull-request**  
Enables textarea/textfield to handle null as a value. This also updates the type for the type-prop for the textfield. 

**Solving issue**  
Fixes: -

**How to test**  
1. Go to storybook link below
2. Check in components -> textfield/textarea
3. Check that the components renders with null as a value. 

**Screenshots**  
-

**Additional context**  
-


alexandereneroth (approved):
Code looks good to me, also tested running it 👍

PR #543, opened 2022-10-25
fix(textfield): display flex on sdds-helper to avoid clash between slot and char counter

mJarsater:
**Describe pull-request**  
Changed the sdds-textfield-helper to display flex and justify-content:between to allow for both the slot, error state and char counter to fit on the same row. 

**Solving issue**  
Fixes: #524 [AB#2443](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2443)

**How to test**  
1. Go to storybook link below
2. Add a helper text on textfield, state to error and set a number on the char count. 
3. Check that all of these elements can be on the same row. 

**Screenshots**  
Before:
Screenshot 2022-10-25 at 16 48 43

After:
Screenshot 2022-10-25 at 16 49 07

**Additional context**  
-

PR #560, opened 2022-11-02
fix(textfield): display flex on sdds-helper to avoid clash between slot and char counter

mJarsater:
This PR updates tegel with updates Tegel to solve a bug that was found in 4_10.0.
PR that has been merged in 4_10.0: https://github.com/scania-digital-design-system/sdds/pull/543

**Describe pull-request**  
Changed the sdds-textfield-helper to display flex and justify-content:between to allow for both the slot, error state and char counter to fit on the same row.

**Solving issue**  
Fixes: -

**How to test**  
 1. Go to storybook link below
 2. Add a helper text on textfield, state to error and set a number on the char count.
 3. Check that all of these elements can be on the same row.

**Screenshots**  
Before: 

After:

**Additional context**  
-

### DIFF
--- a/tegel/src/components/textfield/textfield.scss
+++ b/tegel/src/components/textfield/textfield.scss
@@ -266,7 +266,8 @@
 .sdds-textfield-helper {
   font: var(--sdds-detail-05);
   letter-spacing: var(--sdds-detail-05-ls);
-  display: block;
+  display: flex;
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
Thanks for remembering to bring this one in "new" components too ;)

PR #561, opened 2022-11-02
fix(inline-tabs): make inline tabs html compliant

mJarsater:
**Describe pull-request**  
Made the component HTML compliant by making sure the inline tabs use data-name instead of name and aria-disabled="true" instead of disabled. Since the previous attributes isn't part of the global attributes we can't expects it's children to have those elements. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes

**Solving issue**  
Fixes: [AB#2489](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2489)

**How to test**  
1. Check the component in the sandbox and storybook
2. Make sure it's using data-* / aria-hidden attributes
3. Check that it renders and act in the expected way. 

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/components/src/components/tabs/inline-tabs/inline-tabs.stories.js
+++ b/components/src/components/tabs/inline-tabs/inline-tabs.stories.js
@@ -34,7 +34,7 @@ const Template = ({ autoHeight = false }) => `
       And here is a little bit more.
     </div>
 
-    <div default data-name="Tab 2">
+    <div data-default="true" data-name="Tab 2">
       Content for tab 2<br>
       This is just a little content, but the size of the container is based to the tab with the most content.
     </div>
--- a/components/src/components/tabs/inline-tabs/inline-tabs.stories.js
+++ b/components/src/components/tabs/inline-tabs/inline-tabs.stories.js
@@ -25,7 +25,7 @@ const Template = ({ autoHeight = false }) => `
   </style>
 
   <sdds-inline-tabs ${autoHeight ? 'auto-height' : ''}>
-    <div name="Tab very long name">
+    <div data-name="Tab very long name">
       Content for tab 1<br>
       This tabs has a lot of content so this is the one that decides the height of the container if height="auto" is specified on the component.
       <br><br>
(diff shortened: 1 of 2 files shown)


mJarsater:
Another approach for this one would be to have a <sdds-inline-tabs-content> element I guess. There we could specify the attributes we want without getting TS error. But of course it's a breaking change, which the other solution also is.

alexandereneroth:
@mJarsater Realised this change is breaking. I don't think we can publish it in a bugfix release unfortunately, even though the existing behaviour triggers typescript errors, if we publish this here sites might stop working after update.

alexandereneroth (changes requested):
☝️

mJarsater:
> @mJarsater Realised this change is breaking. I don't think we can publish it in a bugfix release unfortunately, even though the existing behaviour triggers typescript errors, if we publish this here sites might stop working after update.

Yeah I know. I just put this one to merge into 4.10.0 for now. But like you said. Obviously we can't add it to a minor/buxfix update.

mJarsater:
@alexandereneroth have a look at the latest changes. Shouldn't that cover all our cases?

alexandereneroth:
> @alexandereneroth have a look at the latest changes. Shouldn't that cover all our cases?

Yep, that looks like it should cover them 👍

alexandereneroth:
> > @mJarsater Realised this change is breaking. I don't think we can publish it in a bugfix release unfortunately, even though the existing behaviour triggers typescript errors, if we publish this here sites might stop working after update.
> 
> Yeah I know. I just put this one to merge into 4.10.0 for now. But like you said. Obviously we can't add it to a minor/buxfix update.

Alright. I think we have to only do these changes in main though, since thats our next major release as far as i know.
Right @theJohnnyMe ?

mJarsater:
> > > @mJarsater Realised this change is breaking. I don't think we can publish it in a bugfix release unfortunately, even though the existing behaviour triggers typescript errors, if we publish this here sites might stop working after update.
> > 
> > 
> > Yeah I know. I just put this one to merge into 4.10.0 for now. But like you said. Obviously we can't add it to a minor/buxfix update.
> 
> Alright. I think we have to only do these changes in main though, since thats our next major release as far as i know. Right @theJohnnyMe ?

But with my other comment. If we cover both cases I don't see a reason for not including it in a non-major release?

alexandereneroth:
> > > > @mJarsater Realised this change is breaking. I don't think we can publish it in a bugfix release unfortunately, even though the existing behaviour triggers typescript errors, if we publish this here sites might stop working after update.
> > > 
> > > 
> > > Yeah I know. I just put this one to merge into 4.10.0 for now. But like you said. Obviously we can't add it to a minor/buxfix update.
> > 
> > 
> > Alright. I think we have to only do these changes in main though, since thats our next major release as far as i know. Right @theJohnnyMe ?
> 
> But with my other comment. If we cover both cases I don't see a reason for not including it in a non-major release?

Ah, sorry, turns out i didnt look properly. You are right 🙂

theJohnnyMe (approved):

theJohnnyMe:
I like we have logic for covering older ways of passing names while introducing a new (and correct) one.
Please, make sure you edit code examples in our cms for this component too so newver users "copy" the correct code from there too. 

Also, this can be part of new code too, so please juct copy it in Tegel package as well.

PR #563, opened 2022-11-03
fix(banner): add sdds-icon instead of hard-coded SVGs, default value fix

theJohnnyMe:
**Describe pull-request**  
Add sdds-icon instead of hard-coded SVGs, default value fix

**Solving issue**  
Fixes: Returning and fixing "TODO/FIXME comments after link and icon component are migrated"

**How to test**  
1. Check if banner component looks the same as in [main branch](https://tegel-storybook.netlify.app/?path=/docs/components-banner--default).
2. Check if type and icon Control have "default" column filled with default value

### DIFF
--- a/tegel/src/components/banner/banner.stories.ts
+++ b/tegel/src/components/banner/banner.stories.ts
@@ -19,6 +19,9 @@ export default {
       control: {
         type: 'radio',
       },
+      table: {
+        defaultValue: { summary: 'default' },
+      },
     },
     header: {
       name: 'Header',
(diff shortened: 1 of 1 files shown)


tegel/src/components/banner/banner.stories.ts
```diff
@@ -43,38 +46,23 @@ export default {
     },
     prefix: {
       name: 'Show icon',
-      description: 'Icon to display in link section',
+      description: 'Icon to display in link section. For default type "truck" is preselected, but it is up to user needs to select appropriate icon.',
```
alexandereneroth, line 49:
```suggestion
      description: 'If an icon should be displayed. For type default the truck icon is used in this example, but it should be changed to suit your needs.',
```
WDYT?

alexandereneroth (approved):
Nice 🍮

tegel/src/components/banner/banner.stories.ts
theJohnnyMe: Thanks a lot buddy for suggestion <3

PR #565, opened 2022-11-03
feat(code example): remove code example component and depedencies

albinscania:
Code example component was removed, and all it's depedencies from package.json.

**Solving issue**  
Fixes: #2471

**How to test**  
Check out the repo and make sure the code component (and it's deps. like higlight.js are removed). Might need an npm install for this.


alexandereneroth (approved):
Looks good to me 👍

PR #572, opened 2022-11-07
feat(typography): migrate typography stories

mJarsater:
**Describe pull-request**  
Migration of the typography stories.

**Solving issue**  
Fixes: [AB#2480](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2480)

**How to test**  
1. Go to storybook link below
2. Check in Foundations -> Typography
3. Check that the correct font styles are being applied. 

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/tegel/package.json
+++ b/tegel/package.json
@@ -46,6 +46,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.11.6",
+    "@scania/typography": "^1.2.1",
     "@stencil/core": "^2.13.0",
     "concurrently": "^7.4.0",
     "html-format": "^1.0.2"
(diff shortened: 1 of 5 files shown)


tegel/package.json
```diff
@@ -46,6 +46,7 @@
   },
   "dependencies": {
     "@popperjs/core": "^2.11.6",
+    "@scania/typography": "^1.2.1",
```
theJohnnyMe, line 49: Yep, you needed this one in order to get proper class names in, right?

Please, make a ticket for typography under Fresh Stencil & Storybook to come back to it. Maybe we can include other scss files in global and avoid installing packages. Just put a ticket so we discuss it.

theJohnnyMe (approved):
Looks good to me ;)

tegel/package.json
mJarsater: Yeah exactly. So this way we are using the actual package instead of creating classnames just for storybook.

I'll create a ticket for checking which approach is better 😄

PR #574, opened 2022-11-08
fix(inline-tabs): make inline-tabs valid html in tegel

mJarsater:
**Describe pull-request**  
Makes inline-tabs html compliant, but keeps an option of using old way aswell.

**Solving issue**  
Fixes: [AB#2493](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2493)

**How to test**  
2. Check in Components -> inline-tabs
3. Check that the component uses valid HTML.

**Screenshots**  
-

**Additional context**  
-


theJohnnyMe (approved):
Thanks for remembering to bring this one back to the "Tegel" code too!

PR #577, opened 2022-11-08
build: enforce node 16 to prevent dependency issues

alexandereneroth:
**Describe pull-request**  
This PR enforces the node version to be version 16 during npm install to prevent dependency issues.

**How to test**  
Run npm install with any other node version than 16.

**Screenshots**  
With this PR, if you use the wrong node version it will look something like this:
Screenshot 2022-11-08 at 11 04 26


theJohnnyMe (approved):
Netlify built it, we will test Amplify down the road too. Good suggestion ;)

Issue #440, opened 2022-09-09
Dropdown / Label-inside bug


alexlioliopoulos:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have made sure that the issue comes from SDDS packages by removing SDDS from my project as a test.
- [X] I have followed the installation guide.

### Package

@scania/components

### Package versions

4,7

### Browser

Chrome

### Framework

Angular

### Reproduction steps

![image](https://user-images.githubusercontent.com/76053160/189371760-ffd04c61-ee29-40f3-ad25-82f1f61b6d8f.png)

Dropdown with label inside is not rendering the placeholder text.

### Code example

<div>....

### Screenshots

Add/Copy screenshot here

### Expected behaviour

Component x should be able to...

### Console errors

Add warning/error message

[AB#2321](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2321)

PR #584, opened 2022-11-09
fix(dropdown): inside label full size when no option is selected

EmelieLitwin:
**Same as previous PR but made to old main.**

**Describe pull-request**
Dropdown with label inside is not rendering the label properly when no option is selected. Not changed for new Tegel components yet

**Solving issue**
Github issue https://github.com/scania-digital-design-system/sdds/issues/440 and Azure ticket [AB#2321](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2321)

**How to test**

- Go to storybook
- Check dropdown label inside and see if it looks correct


alexandereneroth (changes requested):
When i use label position inside and manually select an option it looks like this for me:
Screenshot 2022-11-10 at 11 01 12

(tested in our storybook)

theJohnnyMe:
> When i use label position inside and manually select an option it looks like this for me: Screenshot 2022-11-10 at 11 01 12
> 
> (tested in our storybook)

Great catch!

theJohnnyMe (changes requested):
+1 on Alex finding

theJohnnyMe (approved):

PR #587, opened 2022-11-09
style: make linting and formatting more useful

alexandereneroth:
Use the project wide eslint, prettier, and tsconfig for the tegel package for consistency and less
confusion. 
Also adjusts linting rules to work with prettier, and typescript.

**Solving issue**  
Fixes: [AB#2507](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2507)


albinscania (approved):

theJohnnyMe (approved):
All in all, it looks good to me. One thing I would be good to have is scripts in package.json for the "check" and "write" options of both prettier and eslint, but it can be a part of another PR.
Also, when doing so, please double-check that husky/pre-commit hooks have some of those in as well for this tegel folder.

PR #590, opened 2022-11-11
build: fix build failing on netlify

alexandereneroth:
I had removed prettier in the tegel package, thinking that we only used that for linting while developing, when we're in-fact using it to format code examples in storybook as well.
This PR re-adds prettier to the tegel packages dependencies.


mJarsater (approved):

theJohnnyMe (approved):
Great! Thanks for fixing it quickly ;)

PR #595, opened 2022-11-14
fix(dropdown): label bugs in new main

EmelieLitwin:
**Describe pull-request**
Dropdown with label inside is not rendering the label properly
image

**How to test**

- Go to storybook
- Check dropdown label inside and see if it looks correct


theJohnnyMe (approved):
LGTM!

PR #599, opened 2022-11-15
fix/textfield state count issues

mJarsater:
**Describe pull-request**  
If a user selected a char count without a state the counter would render at the wrong place. Also if the value of the textfield/textarea is null nothing would render in the char counter, this adds a "0" in that case.

**Solving issue**  
Fixes: [AB#2574](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2574)

**How to test**  
1. Checkout branch and run storybook.
2. Check in Components -> Textfield/Textarea
3. Add a char counter with and without a state and see that it renders correctly. 

**Screenshots**  
-

**Additional context**  
-


theJohnnyMe:
@mJarsater  - It looks good buddy. Issues I could only reproduce for text field component.
Anyway, I have merged changes for main_4.10.0 so now you can use "main" branch to checkout and to merge back to. No need for 4.10.0 anymore. 
Please change it and I will approve asap.

theJohnnyMe (approved):

PR #605, opened 2022-11-16
feat(storybook): google analytics addon

theJohnnyMe:
**Describe pull-request**  
Adding Google Analytics to the Storybook in Netlify

**Solving issue**  

Fixes: [AB#2590](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2590)

**How to test**  
1. I think we just have to release it and monitor the dashboard in Google Analytics.


tegel/.storybook/manager.js
```diff
@@ -1,5 +1,8 @@
 // .storybook/manager.js
 
+window.STORYBOOK_GA_ID = 'G-T4S5M709VL';
+window.STORYBOOK_REACT_GA_OPTIONS = {};
```
mJarsater, line 4: Should we maybe have this as an environment variable instead?

theJohnnyMe: I was following the instructions from [here](https://www.npmjs.com/package/@storybook/addon-google-analytics) so I just added it where it is suggested and I would like to test it, what do you say?

mJarsater: If we do it like this we are adding our secret/ID to the git history of this repo. So I would suggest we "change" the secret/ID and switch it over to a .env-file after we merge and test this. Otherwise our secret/ID is out in the open.

mJarsater: window.STORYBOOK_GA_ID = process.env.STORYBOOK_GA_ID;

tegel/.storybook/manager.js
```diff
@@ -1,5 +1,8 @@
 // .storybook/manager.js
 
+window.STORYBOOK_GA_ID = 'G-T4S5M709VL';
```
mJarsater, line 3:
```suggestion
window.STORYBOOK_GA_ID = process.env.STORYBOOK_GA_ID;
```

mJarsater (approved):

PR #609, opened 2022-11-16
fix(header): refactored header stories to include the js in the code …

mJarsater:
Under construction 👷🏼 

**Describe pull-request**  
-

**Solving issue**  
Fixes: #

**How to test**  
1. Go to...
2. Check in...
3. Run ...

**Screenshots**  
-

**Additional context**  
-


victorscania:
Most of it seems to work fine, but something is up with the "with searchbar" story.
I sometimes get an error "Void elements do not have end tags "input" (105:9)" sometimes it shows nothing.

victorscania (commented):
Most of it seems to work fine, but something is up with the "with searchbar" story.
I sometimes get an error "Void elements do not have end tags "input" (105:9)" sometimes it shows nothing.

alexandereneroth (approved):

PR #630, opened 2022-11-22
feat(datatable): use required prop for id instead of id attribute

alexandereneroth:
Since the id attribute was required for datatable to work, this PR replaces it with a tableId property in the web-component. The property is automatically initialised to a random string so the user doesn't have to set it themselves.

Screenshot 2022-11-22 at 11 42 38

### DIFF
--- a/tegel/src/components/data-table/table-body-cell/table-body-cell.tsx
+++ b/tegel/src/components/data-table/table-body-cell/table-body-cell.tsx
@@ -34,7 +34,7 @@ export class TableBodyCell {
 
   @State() whiteBackground: boolean = false;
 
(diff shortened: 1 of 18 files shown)


mJarsater (approved):

theJohnnyMe (approved):
Please, just add an extra message to the tableID prop that we talked about. Something like: "table will automatically assign ID for its internal communication, however, we recommend a user to set its own value in case one is using custom events."

alexandereneroth:
I added this under the notes tab. Unfortunately the column gets squished. @theJohnnyMe 

Screenshot 2022-11-25 at 15 09 13

PR #637, opened 2022-11-23
docs(all): removes alphabetic sorting of controls.

mJarsater:
**Describe pull-request**  
This removes the alphabetic sorting of controls, so they are presented as intended.

**Solving issue**  
Fixes: [AB#2679](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2679)

**How to test**  
2. Check in the controls of any component.
3. Make sure they are ordered in the same way they are represented in the default export. 

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/tegel/.storybook/preview.js
+++ b/tegel/.storybook/preview.js
@@ -75,7 +75,6 @@ export const parameters = {
       date: /Date$/,
     },
     expanded: true,
-    sort: 'alpha',
   },
   viewMode: 'docs',
   previewTabs: {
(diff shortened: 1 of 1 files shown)


EmelieLitwin (approved):
Great! 
I do think this means we need to go through controls on every story in storybook and decide what order they should be in. Make some general guidelines. For example type, variant and size should be higher up and margin stuff should be further down the list imo

PR #665, opened 2022-11-29
docs: add storybook-addon-html

alexandereneroth:
Just tried storybook-addon-html quickly, which works better for displaying code.

However, it hydrates the components before displaying their html, which makes much code not copy-pastable.

Related to [AB#2639](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2639)

**Screenshots**  

storybook-addon-html

Screenshot 2022-11-29 at 11 10 23

### DIFF
--- a/tegel/.storybook/main.js
+++ b/tegel/.storybook/main.js
@@ -6,6 +6,7 @@ module.exports = {
     '@storybook/addon-interactions',
     '@storybook/addon-notes/register',
     'storybook-dark-mode',
+    '@whitespace/storybook-addon-html',
   ],
   framework: '@storybook/html',
 };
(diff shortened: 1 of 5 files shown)


mJarsater:
Seems like this one had some build issues related to the spinner?

mJarsater:
I don't know if I think this is any better than just using the docs tab like we are today?

EmelieLitwin (approved):
Can't remember if we made a decision on this on Wednesday but I like it :)

alexandereneroth:
I personally don't think we should merge this, because it doesn't work well for many web-components, since it shows the hydrated HTML.

mJarsater:
> I personally don't think we should merge this, because it doesn't work well for many web-components, since it shows the hydrated HTML. Unsure how to fix this for now.

Agree 👍

Issue #550, opened 2022-10-27
Dropdown label and text placement


alexlioliopoulos:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have made sure that the issue comes from SDDS packages by removing SDDS from my project as a test.
- [X] I have followed the installation guide.

### Package

@scania/tegel-old

### Package versions

Storybook

### Browser

Chrome

### Framework

Other

### Reproduction steps

Storybook

![image](https://user-images.githubusercontent.com/76053160/198352401-b7557137-3498-4f75-aadf-70258af3915f.png)

Documentation

![image](https://user-images.githubusercontent.com/76053160/198352467-1b11a748-880a-4d49-ae8d-02c46146e134.png)

The padding-bottom of the dropdown large is 20px in Storybook, but is designed to be 16px. The padding top is also wrong, but I cannot see the actual pixel number from the inspection tool.

### Code example

<div>....

### Screenshots

Add/Copy screenshot here

### Expected behaviour

Component x should be able to...

### Console errors

Add warning/error message

[AB#2463](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2463)

mJarsater:
Fixed in https://github.com/scania-digital-design-system/sdds/releases/tag/%40scania%2Fcomponent%404.10.4

PR #669, opened 2022-11-30
fix(dropdown): label placement and padding

mJarsater:
**Describe pull-request**  
Changed the bottom padding for the button if the label is inside, also refactored the positioning of the actual inside label. Fixes for both components and tegel.

***NOTE***
I don't know if this is something we want to address when we refactor this components instead of these fixes?

**Solving issue**  
Fixes: [AB#2463](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2463) #550

**How to test**  
2. Check in Components -> Dropdown
3. Check the placement of the inside label and bottom padding of the button when there is an inside label.

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/components/src/components/dropdown/dropdown.scss
+++ b/components/src/components/dropdown/dropdown.scss
@@ -1,15 +1,15 @@
-@import "@scania/typography/dist/scss/mixins";
-@import "@scania/typography/dist/scss/tokens";
-@import "./dropdown-core";
-@import "../../helpers/components-shared.scss";
-@import "../checkbox/checkbox.scss";
+@import '@scania/typography/dist/scss/mixins';
+@import '@scania/typography/dist/scss/tokens';
+@import './dropdown-core';
+@import '../../helpers/components-shared.scss';
+@import '../checkbox/checkbox.scss';
 
 :host,
 ::slotted(*) {
   --sdds-link: var(--sdds-grey-958);
   --sdds-link-text-decoration: none;
 
-  @include type-style("detail-02");
+  @include type-style('detail-02');
 
   color: var(--sdds-grey-958);
   box-sizing: border-box;
@@ -51,7 +51,13 @@
     align-items: center;
 
     &.sdds-dropdown-toggle-label-inside {
-      padding-bottom: 16px;
+      &-lg {
+        padding-bottom: 16px;
+      }
+
+      &-md {
+        padding-bottom: 12px;
+      }
     }
   }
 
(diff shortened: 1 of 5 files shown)


EmelieLitwin:
The inside label specs are hard to review but it looks good! 👍 
I think the label and the chevron are grey-958 instead of grey-700, maybe that could be fixed in this pr?

tegel/src/components/dropdown/dropdown.scss
```diff
@@ -56,6 +56,15 @@
     display: flex;
     align-items: center;
 
+    &.sdds-dropdown-toggle-label-inside {
+      &-lg {
+        padding-bottom: 16px;
+      }
+
+      &-md {
+        padding-bottom: 12px;
```
theJohnnyMe, line 65:
```suggestion
        padding-bottom: 16px;
```

components/src/components/dropdown/dropdown.scss
```diff
@@ -49,6 +49,16 @@
...
     display: flex;
     align-items: center;
+
+    &.sdds-dropdown-toggle-label-inside {
+      &-lg {
+        padding-bottom: 16px;
+      }
+
+      &-md {
+        padding-bottom: 12px;
```
theJohnnyMe, line 59:
```suggestion
        padding-bottom: 16px;
```

theJohnnyMe (changes requested):
When there is no label inside and the component size is "Medium", the text is not in the vertical center. I tested by changing the values of padding to 16px and it looks like one in the Figma file. Please check if that would work for you.

mJarsater:
> When there is no label inside and the component size is "Medium", the text is not in the vertical center. I tested by changing the values of padding to 16px and it looks like one in the Figma file. Please check if that would work for you.

Nice catch! I updated the PR and moved all the padding styling from our dropdown-core.scss to dropdown.scss in order to handle sizes better. Please give it a look 😄

mJarsater:
> @mJarsater - Please double-check the native story, it seems that one needs an update with sizing classes now.

Should be fixed now. Still pondering if we should release this as a fix or just skip it and fix it once we refactor the component?

theJohnnyMe (approved):
LGTM

theJohnnyMe:
> > @mJarsater - Please double-check the native story, it seems that one needs an update with sizing classes now.
> 
> Should be fixed now. Still pondering if we should release this as a fix or just skip it and fix it once we refactor the component?

If there is anything left, please add it as a ticket so we do not forget about it.

PR #670, opened 2022-11-30
feat(): add design and a11y addons to sb

EmelieLitwin:
Added storybook add-on for Figma, a11y, and screen reader.

Table is not linked because all the sub-components are different links right now. There will probably be a reorganization in figma next year that are going to make it easier.


mJarsater (approved):
LGTM!

alexandereneroth (approved):
LGTM 👍

PR #677, opened 2022-12-01
fix(textfield): changed the nominwidth prop to noMinWidth

mJarsater:
**Describe pull-request**  
Changed the prop name to conform with convention and refactored story to include a demo-wrapper.

BREAKING CHANGE: Changed the prop name from nominwidth to no-min-width

**Solving issue**  
Fixes: [AB#2743](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2743)

**How to test**  
1. Go to storybook link below
2. Check in Components -> Textfield
3. Check that the no-min-width prop works as intended. Also check the migration.mdx.

**Screenshots**  
-

**Additional context**  
-


theJohnnyMe (changes requested):
Instead of "radio" as the Storybook control type, use boolean as that better reflects status, then we do not have to say default or no min width, then it becomes just on or off...

theJohnnyMe (approved):
LGTM! ;)

EmelieLitwin (approved):
LGTM! 🙂

PR #678, opened 2022-12-01
feat(accordion): api for panel padding reset

alexandereneroth:
**Describe pull-request**  

This pr 
- renames paddingReset to equalPadding
- updates documentation descriptions
- adds the shadow parts api to the `accordion-item`

Related to [AB#2640](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2640)


tegel/src/components/accordion/accordion.scss
```diff
@@ -71,10 +72,6 @@
       margin: 0;
       padding: 0;
     }
```
alexandereneroth, line 74:
```suggestion
    }
    
    &--padding-reset {
      padding-right: var(--sdds-spacing-element-16);
    }
```

mJarsater:
I guess the question is if this makes the api easier to use or if if complicates things? Maybe a first step would be to just rename the prop to something more inline with other component props.

alexandereneroth:
> I guess the question is if this makes the api easier to use or if if complicates things? Maybe a first step would be to just rename the prop to something more inline with other component props.

I kept the old `padding-reset` attribute in this PR, just added shadow parts and a storybook example how to use it.

Not sure how we should rename it, i guess `less-padding-right` is what it does. Or maybe `symmetrical-padding`, `equal-padding` or `panel-padding="ignore-icon"`. Just throwing out some idéas. 🤔
I couldn't come up with something that was good enough to make me feel it warrants a breaking change.

I think this might be a case where "less is more" and we should just remove the attribute and then users can do what they want with the shadow parts API instead.

If not that then i suggest something like `panel-padding="ignore-icon"`, and then we have room to also use it for something like `panel-padding="none"` to have no padding at all if that is requested in the future.

What do you think?

mJarsater:
> > I guess the question is if this makes the api easier to use or if if complicates things? Maybe a first step would be to just rename the prop to something more inline with other component props.
> 
> I kept the old `padding-reset` attribute in this PR, just added shadow parts and a storybook example how to use it.
> 
> Not sure how we should rename it, i guess `less-padding-right` is what it does. Or maybe `symmetrical-padding`, `equal-padding` or `panel-padding="ignore-icon"`. Just throwing out some idéas. 🤔 I couldn't come up with something that was good enough to make me feel it warrants a breaking change.
> 
> I think this might be a case where "less is more" and we should just remove the attribute and then users can do what they want with the shadow parts API instead.
> 
> If not that then i suggest something like `panel-padding="ignore-icon"`, and then we have room to also use it for something like `panel-padding="none"` to have no padding at all if that is requested in the future.
> 
> What do you think?

I just think added the shadow parts might be confusing since we aren't using it anywhere else. I agree that the naming is tricky and it might not warrant a breaking change, but maybe we shouldn't change it then? `paddingReset` might be good for now and if we run into the same issue in other components we can have a look at naming again to have them aligned.

theJohnnyMe:
I agree with Martin on this one. Let's just rename paddingReset to something more intuitive, I like the idea of padding symmetrical-padding or equal-x-padding. It is good to know about this "part" selector, but I also feel it would add to the confusion that already exists. 

I wish that we just have one padding in the design and let the users add margin or padding in the content they add instead of us introducing toggles for padding, but what can you do? 😕

alexandereneroth:
> > > I guess the question is if this makes the api easier to use or if if complicates things? Maybe a first step would be to just rename the prop to something more inline with other component props.
> > 
> > 
> > I kept the old `padding-reset` attribute in this PR, just added shadow parts and a storybook example how to use it.
> > Not sure how we should rename it, i guess `less-padding-right` is what it does. Or maybe `symmetrical-padding`, `equal-padding` or `panel-padding="ignore-icon"`. Just throwing out some idéas. 🤔 I couldn't come up with something that was good enough to make me feel it warrants a breaking change.
> > I think this might be a case where "less is more" and we should just remove the attribute and then users can do what they want with the shadow parts API instead.
> > If not that then i suggest something like `panel-padding="ignore-icon"`, and then we have room to also use it for something like `panel-padding="none"` to have no padding at all if that is requested in the future.
> > What do you think?
> 
> I just think added the shadow parts might be confusing since we aren't using it anywhere else. I agree that the naming is tricky and it might not warrant a breaking change, but maybe we shouldn't change it then? `paddingReset` might be good for now and if we run into the same issue in other components we can have a look at naming again to have them aligned.



> I agree with Martin on this one. Let's just rename paddingReset to something more intuitive, I like the idea of padding symmetrical-padding or equal-x-padding. It is good to know about this "part" selector, but I also feel it would add to the confusion that already exists.
> 
> I wish that we just have one padding in the design and let the users add margin or padding in the content they add instead of us introducing toggles for padding, but what can you do? 😕

Removed the shadow parts from the example. Do you think i should remove the parts themselves as well?

"kill your darlings" my teacher told me.. seems i haven't learned that lesson yet.

alexandereneroth:
Went with equalPadding. 
Although I kind of feel like we should just keep paddingReset and close this.

theJohnnyMe:
Kill it.  📖🔫  There is more joy in coding elsewhere than in this task.

theJohnnyMe:
Closing this one.

PR #684, opened 2022-12-02
fix(message): refactored class names to single dash and updated scss and story

mJarsater:
**Describe pull-request**  
Refactored class names to use single dash instead of __ and --. Also moved color of icon from external style to component css

BREAKING CHANGE: Class names that contained __ or -- was changed to single dash(-)

**Solving issue**  
Fixes: AB#2701

**How to test**  
2. Check in Components -> Message
3. Check that the styling of the component works as intended.
4. Check the changes in the migration.mdx

**Screenshots**  
-

**Additional context**  
-


theJohnnyMe (approved):
LGTM

Issue #589, opened 2022-11-10
[Feature Request]: Add disabled option to sdds-dropdown-option element


toliv3:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md)
- [X] I have checked the lastest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Sometimes we wish to show options that aren't actually available for the user to select them in the current context.
Using the native option element one can add a disable option to still show the option to the user, but make it unavailable for selection, this request is to add a similar feature to the sdds-dropdown-option.

### Background

Some applications use the native select/option disable option to show the user that one option is not available in the current context. We need similar funcionality if we are to move to sdds.

### Idea

Add a disabled option, the item would still be shown in the dropdown, but the element would have a different styling and clicking it would not trigger the selected event.

### Contact information

thiago.oliveira@scania.com

[AB#2549](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2549)

mJarsater:
Added in https://github.com/scania-digital-design-system/sdds/releases/tag/%40scania%2Fcomponent%404.10.4

PR #689, opened 2022-12-02
feat(dropdown-option): adding disabled state

theJohnnyMe:
**Describe pull-request**  
Adding disabled state to dropdown option

**Solving issue**  
Fixes: [AB#2549](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2549) #589 

**How to test**  
1. Go to the dropdown component
2. In both the default web component and multi-select one, there is one item that is disabled
3. Try to select it, it should not be possible

**Additional context**  
Currently, this fix is not involving filtering type of dropdown. 
The feature is done for both the component and Tegel package
The design of a disabled state has to be verified by one of our designers.

### DIFF
--- a/components/src/components.d.ts
+++ b/components/src/components.d.ts
@@ -225,6 +225,10 @@ export namespace Components {
     }
     interface SddsDropdownOption {
         /**
+          * Sets option to disabled state if true
+         */
+        "disabled": boolean;
+        /**
           * Selected set to true if selected
          */
         "selected": boolean;
@@ -1059,6 +1063,10 @@ declare namespace LocalJSX {
         "state"?: string;
     }
     interface SddsDropdownOption {
+        /**
+          * Sets option to disabled state if true
+         */
+        "disabled"?: boolean;
         "onSelectOption"?: (event: CustomEvent<any>) => void;
         /**
           * Selected set to true if selected
--- a/components/src/components/dropdown/dropdown-option.tsx
+++ b/components/src/components/dropdown/dropdown-option.tsx
@@ -24,6 +24,9 @@ export class DropdownOption {
   /** Selected set to true if selected */
   @Prop() selected: boolean = false;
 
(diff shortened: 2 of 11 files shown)


mJarsater (changes requested):
LGTM other than double dash on 'sdds-dropdown-option--disabled'. I would suggest we change it to single dash 'sdds-dropdown-option-disabled' to conform with convention.

matyldamc:
The functionality works, but I think there might be some visual issues between the multi-select web component, and what's in the Figma UI library. Screenshot attached showing what's in Storybook, and Figma.
Screenshot 2022-12-06 at 09 11 30

theJohnnyMe:
> LGTM other than double dash on 'sdds-dropdown-option--disabled'. I would suggest we change it to single dash 'sdds-dropdown-option-disabled' to conform with convention.

Great catch! It should be fixed now!

theJohnnyMe:
> The functionality works, but I think there might be some visual issues between the multi-select web component, and what's in the Figma UI library. Screenshot attached showing what's in Storybook, and Figma. Screenshot 2022-12-06 at 09 11 30

I think you accidentally went into "dark mode" where we have not yet updated variables to correct one - dark mode is still under development. Please check in "light mode", make sure this icon is appearing:    
Screenshot 2022-12-06 at 11 28 50

PS: Charlotte and I had sync on colors to so I have updated the disabled test to grey-400, while the background color stays the same.

matyldamc:
@theJohnnyMe ok good to know! In that case, the only thing that's off is the background color of the selected-disabled option should be white :)
Screenshot 2022-12-06 at 11 37 13

matyldamc (approved):
@theJohnnyMe disregard my comment about the bg state. Didn't read your entire post :) Approved!

mJarsater (approved):
LGTM

PR #700, opened 2022-12-06
feat(modal): refactored how we open the modal and removed prop selector

mJarsater:
**Describe pull-request**  
Refactored how we open the modal and removed prop selector, updated the story and added to migration docs

BREAKING CHANGE: Removed the selector prop and added prop for open.

**Solving issue**  
Fixes: [AB#2739](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2739)

**How to test**  
2. Check in Components -> Modal -> Web component
3. Check that the modal works as expected.
4. Check the added chapter in migration docs.

**Screenshots** 
-

**Additional context** 
-

### DIFF
--- a/tegel/src/components/modal/modal.tsx
+++ b/tegel/src/components/modal/modal.tsx
@@ -26,9 +26,8 @@ export class Modal {
     this.open = true;
   }
 
-  dismissModal() {
+  setDissmissButtons() {
     const nodes = this.el.querySelectorAll('[data-dismiss-modal]');
-
     nodes.forEach((el) => {
       el.addEventListener('click', () => {
         this.open = false;
@@ -45,6 +44,10 @@ export class Modal {
     ) {
       this.open = false;
     }
+  }
+
+  componentDidRender() {
+    this.setDissmissButtons();
   }
 
   render() {
--- a/tegel/src/components/modal/modal.tsx
+++ b/tegel/src/components/modal/modal.tsx
@@ -1,4 +1,4 @@
-import { Component, h, Listen, Host, Prop, State, Element, Watch, Method } from '@stencil/core';
+import { Component, h, Host, Prop, Element, Method } from '@stencil/core';
 
 @Component({
   tag: 'sdds-modal',
@@ -6,9 +6,6 @@ import { Component, h, Listen, Host, Prop, State, Element, Watch, Method } from 
   shadow: true,
 })
 export class Modal {
-  /** Target selector that triggers opening of modal, for example button with id="btn1", then selector is "#btn1" */
-  @Prop() selector: string = '';
-
   /** Disables closing modal on clicking on overlay area. */
   @Prop() prevent: boolean = false;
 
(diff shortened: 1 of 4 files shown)


tegel/src/components/modal/modal.tsx
```diff
@@ -18,69 +15,51 @@ export class Modal {
...
   }
 
   render() {
     return (
-      <Host class={`sdds-modal-backdrop ${this.show ? 'show' : 'hide'}`}>
-        <div class={`sdds-modal ${this.actions ? `sdds-modal__actions-${this.actions}` : ''} ${this.size ? `sdds-modal-${this.size}` : ''} `}>
+      <Host
+        onClick={(event) => {
+          this.handleClick(event);
+        }}
```
alexandereneroth, line 55:
```suggestion
        onClick={this.handleClick}
```
Did you try this?

mJarsater: Yeah and that works fine, I'll commit it!

tegel/src/stories/Migration/Migration.stories.mdx
```diff
@@ -788,3 +788,57 @@ Change all instances the toast dismiss button like shown below:
...
+
+[Web component](/docs/components-modal--web-component)
+
+#### Refactored how the modal is opened and removed dependacy to trigger
+The modal component has previously been opened via a trigger that was added on a button,
+which would be the controler of the button. This has now been refactored and the modal is
+now opened based on its own boolean prop, called `open`. This change is done in order to
+make the component more inline with how these sort of thing are generally done. We also
+dont want to assume that is is a button that controls the open/closed state of the modal
+and this approach gives you more freedom on how you open your modals.
```
alexandereneroth, line 802:
```suggestion
The modal component was previously opened via a selector that was added on a button,
which would be the controller of the modal. This has now been changed and the modal is
now opened based on its own boolean prop, called `open`. This change is done in order to
make the component more in line with how these sort of thing are generally done. We also
don't want to assume that the open/closed state of the modal is controlled by a button. This approach gives you more freedom on how you open your modals.
```
Reformulated some things, and fixed some typos

alexandereneroth (changes requested):
I like 👍 But suggested some changes to the migration file text 👇

tegel/src/components/modal/modal.tsx
```diff
@@ -18,69 +15,49 @@ export class Modal {
...
 
   /** Shows the modal  */
   @Method()
   async openModal() {
-    this.show = true;
-  }
-
-  componentDidLoad() {
-    const targets = document.querySelectorAll(this.selector);
-    this.dismissModal();
```
theJohnnyMe, line 34: I think we need dismissModal here in order for buttons with the `data-dismiss-modal` attribute to close. 
What do you think?

theJohnnyMe (changes requested):

tegel/src/components/modal/modal.tsx
mJarsater: Do you mean the button within the modal itself or button from outside the modal?

theJohnnyMe: I was led by the story example, where both the delete and cancel buttons inside of the dropdown (slotted ones) have that attribute. I wonder if we run that function in the lifecycle or if we completely remove it and rely only on method/prop for closing?

What do you think?

mJarsater: Ah alright! I guess the question then if whether or not we want the component to work in that way? Since we are opening/closing the modal by toggling the open prop perhaps these buttons should do that as well. But I don't know if they should by default..

theJohnnyMe (approved):
Nice work man!

alexandereneroth (approved):
⭐

PR #702, opened 2022-12-06
docs(): add comments to demo wrappers in storybook

EmelieLitwin:
**Describe pull-request**  
Added comments to demo wrappers in storybook

**Solving issue**  
[AB#2680](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2680)

**How to test**  
Check storybook, ex button

### DIFF
--- a/tegel/src/components/button/button-native.stories.ts
+++ b/tegel/src/components/button/button-native.stories.ts
@@ -160,15 +160,15 @@ const NativeTemplate = ({
   return formatHtmlPreview(
     `
   <style>
+  /* demo-wrapper is for demonstration purposes only*/
     ${
       icon && iconType === 'Native'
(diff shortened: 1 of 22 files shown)


theJohnnyMe (changes requested):
I have a couple of suggestions:

1. Make a 'enter' between </script> or </style> so there is space - it will separate demo script/classes from the main code. It will be easier for those that copy code.
2. Add a comment for <script> tags too, there are couple of components where we have JS for demonstration purposes - makes sense to comment on that code too. (eg. side-menu, native modal...)
3. date-time component has inline width - make a demo-wrapper so we are consistent with the rest of the code.
4. Popover-menu: see if you can group all demo code to the top of the page

PR #710, opened 2022-12-09
fix(typography): refactored px to rem for font-size and no unit for line-height

mJarsater:
**Describe pull-request**  
Refactored both tegel font-vars and typography-package vars to use rem for font-size and unitless for line-height.

**Solving issue**  
Fixes: AB#2830

**How to test**  
-

**Screenshots**  
-

**Additional context**  
-


alexandereneroth:
I thought the old typography package was using rems, is this not the case? (before @theJohnnyMe  changed it some weeks ago)

mJarsater:
> 

If it did it was based on base 4. Which it's not anymore, so i guess this would have to be highlighted.

mJarsater:
> I thought the old typography package was using rems, is this not the case? (before @theJohnnyMe changed it some weeks ago)



> > I thought the old typography package was using rems, is this not the case? (before @theJohnnyMe changed it some weeks ago)
> 
> If it did it was based on base 4. Which it's not anymore, so i guess this would have to be highlighted.

Well I guess this would also mean we change the theme-light to not set base 4px. Which we might not wanna do now. So should this PR only update the font-vars that are **inside** the tegel package, and not poke around in typography package?

PR #720, opened 2022-12-12
fix(date-time): align props date time and introduce default-value

mJarsater:
**Describe pull-request**  
Removed the slots for label/helper and created props for them instead, also introduces the prop
default-value to allow the user to select a default time to be displayed.

**Solving issue**  
Fixes: [AB#2837](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2837)

**How to test**  
2. Check in Components -> Date-time
3. Check that the controls regarding label/helper works as inteded.
4. Try out the new custom/now value for default-value.

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/tegel/src/components/datetime/datetime.tsx
+++ b/tegel/src/components/datetime/datetime.tsx
@@ -15,10 +15,8 @@ export class Datetime {
   /** Value of the input text */
   @Prop({ reflect: true }) value = '';
 
-  /** Default value of the component. Needs to match the format of the type.
-    Use 'now' to get the current date/time.
-   */
-  @Prop() defaultValue: string | 'now' | 'none' = 'none';
+  /** Default value of the component. Needs to match the format of the type. */
+  @Prop() defaultValue: string | 'none' = 'none';
 
   /** Set input in disabled state */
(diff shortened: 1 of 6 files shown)


tegel/src/components/datetime/datetime.tsx
```diff
@@ -65,6 +133,7 @@ export class Datetime {
   // Change event isn't a composed:true by default in for input
   handleChange(e): void {
     this.customChange.emit(e);
+    console.log(this.value);
```
mJarsater, line 136:
```suggestion
```

mJarsater:
Question: Is it better to use the value prop to set the default value?

theJohnnyMe:
Props format and defaultValue - are they dependent on each other and if how?
**format:** prop looks like it is not used - but maybe you wanted to discuss it more in the group? 
**defaultValue:** Do we need to specify the format that is expected in the comment of the prop so users know in which format they need to pass the value?

theJohnnyMe (approved):
Nice work! LGTM!

PR #730, opened 2022-12-16
feat(breadcrumb): webcomponent

mJarsater:
**Describe pull-request**  
Web component version of the breadcrumb.

**Solving issue**  
Fixes: [AB#2854](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2854)

**How to test**  
1. Go to storybook link below
2. Check in Components -> Breadcrumb -> WebComponent

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/tegel/src/components/breadcrumb/breadcrumb-component.stories.ts
+++ b/tegel/src/components/breadcrumb/breadcrumb-component.stories.ts
@@ -1,6 +1,11 @@
+import readme from './readme.md';
+import itemReadme from './breadcrumb-item/readme.md';
+import { formatHtmlPreview } from '../../utils/utils';
+
 export default {
   title: 'Components/Breadcrumb',
   parameters: {
+    notes: { 'Breadcrumb': readme, 'Breadcumb Item': itemReadme },
     design: [
       {
         name: 'Figma',
@@ -16,13 +21,16 @@ export default {
   },
 };
 
-const Template = () => `   
+const Template = () =>
+  formatHtmlPreview(
+    `   
       <sdds-breadcrumb>
       <sdds-breadcrumb-item href="#" link-text="A breadcrumb"></sdds-breadcrumb-item>
       <sdds-breadcrumb-item href="#" link-text="Another breadcrumb"></sdds-breadcrumb-item>
         <sdds-breadcrumb-item link-text="Current one" current></sdds-breadcrumb-item>
       </sdds-breadcrumb>
-      `;
+      `,
+  );
 
 export const WebComponent = Template.bind({});
 
--- a/tegel/src/components/breadcrumb/breadcrumb-component.stories.ts
+++ b/tegel/src/components/breadcrumb/breadcrumb-component.stories.ts
@@ -19,7 +19,6 @@ export default {
 const Template = () => `   
       <sdds-breadcrumb>
       <sdds-breadcrumb-item href="#" link-text="A breadcrumb"></sdds-breadcrumb-item>
-      <sdds-breadcrumb-item disabled href="#" link-text="Disabled one"></sdds-breadcrumb-item>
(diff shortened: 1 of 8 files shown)


tegel/src/components/breadcrumb/breadcrumb-item/breadcrumb-item.tsx
```diff
@@ -0,0 +1,34 @@
+import { Component, Host, h, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'sdds-breadcrumb-item',
+  styleUrl: 'breadcrumb-item.scss',
+  shadow: true,
+})
+export class SddsBreadcrumb {
```
mJarsater, line 8:
```suggestion
export class SddsBreadcrumbItem {
```

EmelieLitwin:
Yeay! Making web components 😍 
Some thoughts:
Maybe the breadcrumb could be a nav-element instead of host and have the slots in an ol for WCAG reasons:
https://www.w3.org/WAI/ARIA/apg/example-index/breadcrumb/index.html
I don't know if we want to get a head start on a bit of a11y for new components, but there is also an aria state called aria-current.

mJarsater:
> Yeay! Making web components 😍 Some thoughts: Maybe the breadcrumb could be a nav-element instead of host and have the slots in an ol for WCAG reasons: https://www.w3.org/WAI/ARIA/apg/example-index/breadcrumb/index.html I don't know if we want to get a head start on a bit of a11y for new components, but there is also an aria state called aria-current.

Great idea! I'll update the PR 😄

mJarsater:
@EmelieLitwin updated the PR now. Please have a look :)

EmelieLitwin:
> @EmelieLitwin updated the PR now. Please have a look :)

Nice! :) I was thinking that the disabled breadcrumb in the storybook story might not be needed. It looks a little off and hopefully not a common use case.

I also saw in figma that hovered breadcrumb is supposed to be underlined.

Otherwise, looks good 🥳

tegel/src/components/breadcrumb/breadcrumb-component.stories.ts
```diff
@@ -0,0 +1,29 @@
...
+      {
+        name: 'Link',
+        type: 'link',
+        url: 'https://www.figma.com/file/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=2703%3A4725&t=rVXuTOgTmXPauyHd-1',
+      },
+    ],
+  },
+};
+
+const Template = () => `
```
theJohnnyMe, line 19: Please add formatHTML here.

tegel/src/components/breadcrumb/breadcrumb-component.stories.ts
```diff
@@ -0,0 +1,29 @@
+export default {
```
theJohnnyMe, line 1: Please add the breadcrumb item readme to the "Notes" tab of Storybook.

theJohnnyMe (approved):
Please rename the "Default" story to "Native" just as that is how we named it in other components that exist in both HTML&CSS and WebComponent versions. 

Other than that tiny things, it looks good!

mJarsater:
> Please rename the "Default" story to "Native" just as that is how we named it in other components that exist in both HTML&CSS and WebComponent versions.
> 
> Other than that tiny things, it looks good!

Nice catch!

mJarsater:
@theJohnnyMe please have a look again. Removed the link-text prop and used a slot instead, I think it looks a bit neater 😄

theJohnnyMe:
> @theJohnnyMe please have a look again. Removed the link-text prop and used a slot instead, I think it looks a bit neater 😄

That was a good call really, then component reminds more of the native anchor tag + it enables uses to pass HTML in too, in case they want to have an SVG icon text to text too or something similar.

PR #757, opened 2022-12-30
fix(data-table): add dark/light mode variables

EmelieLitwin:
**Describe pull-request**  
Added variables for dark and light mode to data table. 

**Solving issue**  
Fixes: AB#2925


tegel/src/components/data-table/table-body-row/table-body-row.scss
```diff
@@ -1,14 +1,16 @@
 /* As checkbox is not a web component, we have to import them here in order to have styles available - shadow DOM is a reason */
 @import '../../checkbox/checkbox';
 @import '../../../mixins/box-sizing';
+@import '../data-table-theme-vars.scss';
```
alexandereneroth, line 4: see comment in native-table.scss ☝️

alexandereneroth (changes requested):
See other comments above ☝️ 
I also discovered that the "on white background" storybook toggle does not work anymore.

But nice work!

EmelieLitwin:
> See other comments below 👇 I also discovered that the "on white background" storybook toggle does not work anymore.
> 
> But nice work!

Thank you!
I commented out the on-white controls for now because it's going to be changed primary/secondary soon anyway so I'll do it in another ticket/branch.

EmelieLitwin:
Not correct color on selected row?

alexandereneroth (approved):
Awesome work with all the css variables!

PR #758, opened 2022-12-30
fix(modal): fix the constant overflow on modal

mJarsater:
**Describe pull-request**  
Removes the "after" on the modal toolbar and added padding-bottom to prevent constant overflow.

**Solving issue**  
Fixes: -

**How to test**  
1. Checkout and run start for components package.
2. Check in Components -> Modal
3. Check that modal that should not have overflow (little body text) does not have a scrollbar.

**Screenshots**  
-

**Additional context**  
-

### DIFF
--- a/components/src/components/modal/_modal-core.scss
+++ b/components/src/components/modal/_modal-core.scss
@@ -1,5 +1,5 @@
-@import "@scania/typography/dist/scss/mixins";
-@import "@scania/typography/dist/scss/tokens";
+@import '@scania/typography/dist/scss/mixins';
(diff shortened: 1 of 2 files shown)


theJohnnyMe (approved):
Tested on my local. I had just a little portion of text and the scrollbar is gone, compared to the [prod](https://production.d1g8v8mrkx992n.amplifyapp.com/?path=/story/component-modal--html-and-css&args=size:lg) version. 

LGTM! Nice work!

alexandereneroth (approved):
Nice!

PR #727, opened 2022-12-14
build(tegel): committing correct stencil -v, igoring stories on build

mJarsater:
**Describe pull-request**  
Set stencil -v to 2.20.0, bumped npm version to latest, ignoring stories files on build.

**Solving issue**  
Fixes: [AB#2845](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2845)

**How to test**  
Try running it locally and run a build. 

**Screenshots**  
-

**Additional context**  
-

PR #763, opened 2023-01-03
chore: fix importing md files in stories not working

alexandereneroth:
Importing md files in stories previously gave a typescript error. Now it shouldn't.

**How to test**  

Open the project in your editor, make sure eslint and prettier is installed and configured.
Verify there are no errors in story files that import markdown files.

Verify that npm packages are still excluding the stories.

**Screenshots**  
No more error:
Screenshot 2023-01-03 at 12 01 42

**Additional context**  
PR that introduced issue: #727


alexandereneroth:
Unfortunately this change includes the stories in the distribution again.

The only potential way around it i've found so far is to reintroduce exclude in tsconfig and use the `tsx` file extension for stories instead of `ts` to allow importing md files.

mJarsater (approved):
Looks good to me. We can maybe revisit this when we start a propper work on the build pipeline.

PR #772, opened 2023-01-09
fix(datetime): corrected styling of label (default and disabled)

mJarsater:
**Describe pull-request**  
Corrected styling of label (default and disabled). Also removed css stylning for slots since these are not used.

**Solving issue**  
Fixes: #3023

**How to test**  
1. Go to storybook link below
2. Check in Components -> Datetime
3. Check that the label placement and styling is correct for all states/sizes.

**Screenshots**  

**Additional context**


alexandereneroth (approved):
Nice

PR #778, opened 2023-01-10
Improve popover API for frameworks

alexandereneroth:
The plan was to make the API the same as we have on Modal, however, as I wrote in the Azure ticket:

>Doing it the same as with modal does not work, as popovers need to know the size and position of the element they are "popping over".

>So, to at least make them work better with frameworks (like React), i added a 'referenceEl' prop where you can pass the dom element for the popped over element, in addition to the already existing 'selector' prop.

**Solving issue**  
Fixes: [AB#2962](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2962)


mJarsater:
Maybe add a description of this change to the migration docs?

alexandereneroth:
> Maybe add a description of this change to the migration docs?

Since this is not replacing anything, just adding the option to use a referenceEl prop to pass the popped over element, we don't think it needs an entry in the migration docs.

mJarsater (approved):
🦄

PR #779, opened 2023-01-11
fix(modal): fix body text spacing

theJohnnyMe:
**Describe pull-request**  
Fixes for modal:

- body text cut when sticky position of the action bar is set
- padding on right missing on smaller width screen sizes
- enabling control for body text for easier testing
- dropping extra class from story example

**Solving issue**  
Fixes: [AB#2849](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2849)

**How to test**  
1. Go to the Netlify PR link
2. Add a lot of text to the body and see if text display correctly, no matter sticky/static action bar position
3. Check if padding-right on body-text area looks good on all screens

**Additional context**  
In case you are curios to understand the problem better, try running the current master branch and modal, you will see there is an issue with text being cut off when the sticky action bar is set.

### DIFF
--- a/tegel/src/components/modal/modal-native.stories.tsx
+++ b/tegel/src/components/modal/modal-native.stories.tsx
@@ -104,7 +104,7 @@ const Template = ({ headline, bodyText, size, actions, showModal }) =>
           </button>
         </div>
           <div class="sdds-modal-body">
-            ${bodyText}
+            <p>${bodyText}</p>
           </div>
           <div class="sdds-modal-actions">
               <button class="sdds-btn sdds-btn-primary sdds-btn-md">Save</button>
--- a/tegel/src/components/modal/modal-native.stories.tsx
+++ b/tegel/src/components/modal/modal-native.stories.tsx
@@ -72,7 +72,7 @@ const Template = ({ headline, size, actions, showModal }) =>
     position: relative;
     top: 0;
     left: 0;
-    height: 500px;
+    height: 100vh;
   }
   .demo-styles {
     position: absolute;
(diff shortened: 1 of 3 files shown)


tegel/src/components/modal/modal-native.stories.tsx
```diff
@@ -95,9 +104,7 @@ const Template = ({ headline, size, actions, showModal }) =>
           </button>
         </div>
           <div class="sdds-modal-body">
-            <p>
```
alexandereneroth, line 98: I think we should keep the paragraph element here, it seems like more correct markup. Or at least i think we should add styles so the margins are still correct if you use a paragraph element.

tegel/src/components/modal/modal-native.stories.tsx
```diff
@@ -64,15 +73,15 @@ const sizeLookUp = {
...
-const Template = ({ headline, size, actions, showModal }) =>
+const Template = ({ headline, bodyText, size, actions, showModal }) =>
   formatHtmlPreview(
     `    <style>
   /* demo-wrapper and demo-styles is for demonstration purposes only*/
   .demo-wrapper {
     position: relative;
     top: 0;
     left: 0;
-    height: 500px;
```
alexandereneroth, line 75: Nice!

tegel/src/components/modal/modal-native.stories.tsx
alexandereneroth: I could fix this if you prefer

alexandereneroth (changes requested):

tegel/src/components/modal/modal-native.stories.tsx
theJohnnyMe: Thanks fore review, let me bring it back quickly ;)

theJohnnyMe: @alexandereneroth  - Fixed!

alexandereneroth (approved):
🍨 !

PR #780, opened 2023-01-11
build(lerna): updating version adding filtering option and skipping tegel from start & build

theJohnnyMe:
**Describe pull-request**  
Updating version adding filtering option and skipping tegel from start & build

**Solving issue**  
When we run the start and build command from the root, Tegel is also triggered. I added the Lerna filtering option so it skips the Tegel package. Another one is updating Lerna to the latest version.

**How to test**  
1. Run this branch on your local
2. Run "npm run start" (or build) from sdds level - Tegel should not be triggered anymore.


alexandereneroth (approved):
Nice! Seems to work as intended. Tested start, build, and build tegel locally.

PR #784, opened 2023-01-16
Accordion mode variant rename

EmelieLitwin:
**Describe pull-request**  
Updating mode variant naming

**Solving issue**  
Fixes: [AB#2970](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2970)

### DIFF
--- a/tegel/src/components/accordion/accordion-theme-vars.scss
+++ b/tegel/src/components/accordion/accordion-theme-vars.scss
@@ -2,31 +2,43 @@
 .sdds-theme-light {
   --sdds-accordion-border: var(--sdds-grey-300);
   --sdds-accordion-background: transparent;
-  --sdds-accordion-background-hover: var(--sdds-grey-50);
-  --sdds-accordion-background-hover-on-secondary: var(--sdds-white);
+
+  --sdds-accordion-background-hover-primary: var(--sdds-grey-50);
+  --sdds-accordion-background-hover-secondary: var(--sdds-white);
+  --sdds-accordion-background-hover: var(--sdds-accordion-background-hover-primary);
+
   --sdds-accordion-background-focus: transparent;
   --sdds-accordion-background-active: var(--sdds-grey-100);
   --sdds-accordion-color: var(--sdds-grey-958);
   --sdds-accordion-border-focus: var(--sdds-grey-500);
   --sdds-accordion-color-disabled: var(--sdds-grey-400);
 
+  .sdds-mode-variant-primary {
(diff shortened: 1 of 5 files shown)


charlottelorensson:
The primary version (on white) has grey-50 as preview background in Storybook so the hover state for primary is not visible (grey-50 on a grey-50 background). The secondary works cause then the background should be grey-50.

EmelieLitwin:
> The primary version (on white) has grey-50 as preview background in Storybook so the hover state for primary is not visible (grey-50 on a grey-50 background). The secondary works cause then the background should be grey-50.
> 
> Although, the issue will be reversed if resolved by changing the background to white I guess. Could the background color change as well when switching between primary and secondary?

Yes, there is nothing wrong with the component but I also wish we could make it så the background changed when the user switched mode variant. But I'm not sure how to do this so hat it works with both modes. I can create a ticket for that. There is already a ticket for changing background color I think.

tegel/src/components/accordion/accordion-theme-vars.scss
```diff
@@ -2,31 +2,43 @@
...
   --sdds-accordion-border-focus: var(--sdds-grey-500);
   --sdds-accordion-color-disabled: var(--sdds-grey-800);
 
-  .sdds-on-white-background {
-    --sdds-accordion-background-hover: var(--sdds-accordion-background-hover-on-white);
+   .sdds-mode-variant-primary {
+    --sdds-accordion-background-hover: var(--sdds-accordion-background-hover-primary);
+  }
+  .sdds-mode-variant-secondary {
+    --sdds-accordion-background-hover: var(--sdds-accordion-background-hover-secondary);
```
mJarsater, line 42: I think this change (and I guess it goes for all made-variant changes on the native components) need to be added to the migration docs.

EmelieLitwin: > I think this change (and I guess it goes for all made-variant changes on the native components) need to be added to the migration docs.

Good point! I think I added it for button WC and native but not the others.

mJarsater: Nice! I guess it's really the most important for the native components. Maybe we should create a ticket for that?

EmelieLitwin:
@mJarsater Updated mode variant prop to null. What do you think?

mJarsater (approved):
LTGM 🎸

PR #791, opened 2023-01-19
feat(banner): webcomponent

mJarsater:
**Describe pull-request**  
Webcomponent for banner.

**Solving issue**  
Fixes: AB#3141

**How to test**  
2. Check in Components -> Banner -> Web Component
3. Check that the styling is correct and the controls works as intended.

**Screenshots**  

**Additional context**


EmelieLitwin:
Great job 💃 
Just my personal opinion but maybe there should be an icon as default in storybook. I don't think you are supposed to use it without so it looks a little weird. And error should always have the error icon and info the info icon I think.

I was thinking we might want to have an dismissable prop of some kind. People might want to have a banner that can't be closed. Needs to be a version of that in the ui-kit in that case. What do you think @charlottelorensson ? 

Some suggested aria-tags:
`aria-labelledby`, `aria-describedby`, `aria-hidden`,` aria-live` and `aria-atomic`
Should also have `role='banner'`

EmelieLitwin:
Another a11y thing is that I think we should have the option to have a label for the closing button. "Close X" instead of just "X". But it would also need an update to the UI-kit. What do you think @charlottelorensson?

mJarsater:
> Great job 💃 Just my personal opinion but maybe there should be an icon as default in storybook. I don't think you are supposed to use it without so it looks a little weird. And error should always have the error icon and info the info icon I think.
> 
> I was thinking we might want to have an dismissable prop of some kind. People might want to have a banner that can't be closed. Needs to be a version of that in the ui-kit in that case. What do you think @charlottelorensson ?
> 
> Some suggested aria-tags: `aria-labelledby`, `aria-describedby`, `aria-hidden`,` aria-live` and `aria-atomic` Should also have `role='banner'`

The only thing with the default banner is what icon should be used for the "default"-state? Or should the default state be able to accept any icon. I think that might make the API a bit confusing. Nice note on the A11y stuff!

mJarsater:
> Some suggested aria-tags: `aria-labelledby`, `aria-describedby`, `aria-hidden`,` aria-live` and `aria-atomic` Should also have `role='banner'`

How should we use the labelledby and describedby once in this component?

theJohnnyMe:
> > Great job 💃 Just my personal opinion but maybe there should be an icon as default in storybook. I don't think you are supposed to use it without so it looks a little weird. And error should always have the error icon and info the info icon I think.
> > I was thinking we might want to have an dismissable prop of some kind. People might want to have a banner that can't be closed. Needs to be a version of that in the ui-kit in that case. What do you think @charlottelorensson ?
> > Some suggested aria-tags: `aria-labelledby`, `aria-describedby`, `aria-hidden`,` aria-live` and `aria-atomic` Should also have `role='banner'`
> 
> The only thing with the default banner is what icon should be used for the "default"-state? Or should the default state be able to accept any icon. I think that might make the API a bit confusing. Nice note on the A11y stuff!

I agree with Emelie on icons. The default version of Banner seems to accept different icons, where info and warning have predefined ones. 

@EmelieLitwin  - Good comment on "X" please check it with Charlotte.

charlottelorensson:
The prefix icons are optional for banners in the documentation (https://tegel.scania.com/components/banners) but fixed in Figma (although allowing ppl to choose from the entire icon library in practice) so I'm not sure there. The message component may be used without the close (X) icon and can therefore be persistent but the banners (and the rest of the notification components) should have them as standard (https://tegel.scania.com/patterns/notifications). But maybe there are some use cases where the close icon should be optional in banners, I'm not sure. Do you have a particular one in mind? :)

theJohnnyMe (approved):
Nice work buddy, just two small things to consider before merging: 
1. Please see if you can add "Show icon" control to the WebComponent version too as the Native one has it and the documentation supports that feature.
2. "Persistent" control name could be "Hide X button" or "Disable closing"

theJohnnyMe:
In Error and Information type of Banner it is not possible to hide the icon, please make it possible. 
btw - good thinking on adding "Link target" control.

PR #800, opened 2023-01-24
feat(block): webcomponent

mJarsater:
**Describe pull-request**  
Webcomponent version of Block. Also renamed the native story.

**Solving issue**  
Fixes: [AB#3160](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/3160)

**How to test**  
2. Check in Components -> Block -> Webcomponent
3. Check that the styling is correct.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

**Screenshots**  

**Additional context**

### DIFF
--- a/tegel/src/components/block/sdds-block.stories.tsx
+++ b/tegel/src/components/block/sdds-block.stories.tsx
@@ -29,7 +29,7 @@ export default {
       },
       options: ['Primary', 'Secondary'],
       table: {
-        defaultValue: { summary: 'Primary' },
+        defaultValue: { summary: 'primary' },
       },
     },
   },
--- a/tegel/src/components/block/sdds-block.stories.tsx
+++ b/tegel/src/components/block/sdds-block.stories.tsx
@@ -29,7 +29,7 @@ export default {
       },
       options: ['Inherit from parent', 'Primary', 'Secondary'],
       table: {
-        defaultValue: { summary: 'primary' },
+        defaultValue: { summary: 'Inherit from parent' },
       },
     },
   },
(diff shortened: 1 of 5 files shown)


tegel/src/components/block/sdds-block.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+      {
+        name: 'Link',
+        type: 'link',
+        url: 'https://www.figma.com/file/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=9743%3A24020&t=rVXuTOgTmXPauyHd-1',
+      },
+    ],
+  },
+  argTypes: {
+    type: {
+      name: 'Mode variation',
```
nathalielindqvist, line 24: Should the control be called "Mode variant" instead to match other components?

mJarsater: Absolutely! Nice catch 🎣

tegel/src/components/block/sdds-block.tsx
```diff
@@ -0,0 +1,21 @@
...
+  /** Mode variant of the component, based on current mode. */
+  @Prop() modeVariant: 'primary' | 'secondary' = null;
+
+  render() {
+    return (
+      <Host>
+        <div class={`sdds-block-webcomponent sdds-block-${this.modeVariant}`}>
+          <slot></slot>
+        </div>
+      </Host>
```
theJohnnyMe, line 18: What do you think of putting classes to the Host level?
I am just thinking of shorter code, nothing else.

```suggestion
      <Host class={`sdds-block-webcomponent sdds-block-${this.modeVariant}`}> 
          <slot></slot>      
      </Host>
```

theJohnnyMe: Or the second option is to drop <Host> entirely?

tegel/src/components/block/sdds-block.tsx
```diff
@@ -0,0 +1,21 @@
+import { Component, h, Host, Prop } from '@stencil/core';
+
+@Component({
+  tag: 'sdds-block',
+  styleUrl: 'sdds-block.scss',
+  shadow: false,
```
theJohnnyMe, line 6: Any reason for shadow: false?

mJarsater: Yeah it's because I could find a way to properly use the ::slotted pseudo-selector to only target the correct child block. Might be something I'm missing here though 🤔

tegel/src/components/block/sdds-block.tsx
mJarsater: Good idea! We can drop the host element all together

tegel/src/components/block/sdds-block.tsx
mJarsater: Actually I think there is no way of doing it with the shadowDOM right now. The css would have to look something like:
`::slotted(*){
    //style
   ::slotted(*){
   // inner block style
   }
}`

And the ::slotted can't handle nested styling :/

mJarsater: Found a way! Have a look at the latest commit and see what you think 😄 @theJohnnyMe

mJarsater: This way, the blocks should "scale" indefinitely. The sub block will always be the reversed color of its parent.

tegel/src/components/block/sdds-block.tsx
```diff
@@ -1,14 +1,28 @@
...
+  @Element() host: HostElement;
+
+  children: Array<HTMLSddsBlockElement>;
+
+  connectedCallback() {
+    this.children = Array.from(this.host.children).filter(
+      (item) => item.tagName === 'SDDS-BLOCK',
+    ) as HTMLSddsBlockElement[];
+    this.children.forEach((item) => {
+      item.setAttribute('mode-variant', this.modeVariant === 'primary' ? 'secondary' : 'primary');
```
theJohnnyMe: OK, I was about to say why is this logic saying if primary then it should be secondary, I get it - as you nest one in another you always get a different background. Good idea man, worth sharing with others in dec check in.

theJohnnyMe (approved):
Awesome work! LGTM!

PR #804, opened 2023-01-24
fix(divider): fix height/width bug in sb

EmelieLitwin:
**Describe pull-request**  
Bug in docs tab i storybook when changing between horizontal and vertical divider:
image
Specified 1px height/width in demo-div.

### DIFF
--- a/tegel/src/components/divider/divider.stories.tsx
+++ b/tegel/src/components/divider/divider.stories.tsx
@@ -121,6 +121,7 @@ const Template = ({ type, width, direction, height }) => {
         ? `/* demo-div is for demonstration purposes only*/
         .demo-div {
           height: ${height}px;
+          width: 1px;
         }`
         : ''
     }    ${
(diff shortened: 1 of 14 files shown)


tegel/src/components/divider/divider.stories.tsx
```diff
@@ -121,13 +121,15 @@ const Template = ({ type, width, direction, height }) => {
         ? `/* demo-div is for demonstration purposes only*/
         .demo-div {
           height: ${height}px;
+          width: 1px;
```
nathalielindqvist, line 124: Divider is still a bit thick, looks like there is some padding on the demo-div that interferes:
image

tegel/src/components/divider/divider.stories.tsx
```diff
@@ -121,13 +121,15 @@ const Template = ({ type, width, direction, height }) => {
...
         }`
         : ''
     }    ${
       width
         ? `/* demo-div is for demonstration purposes only*/
             .demo-div {
               width: ${width}px;
+              height: 1px;
```
nathalielindqvist, line 132: Divider is still a bit thick, looks like there is some padding on the demo-div that interferes:
image

nathalielindqvist (approved):
That works! Nice 👍

PR #806, opened 2023-01-25
fix(button): only icon button

mJarsater:
**Describe pull-request**  
Corrected the css selector for the only-icon button. Also removed the prop onlyIcon being set in the story, this prop doesn't exist. It is an internal state that is set based on the text value. The old way would lead to the component not being rendered properly.

**Solving issue**  
Fixes: - Found this issue when I was implementing the button in the banner component.

**How to test**  
2. Check in Components -> Button -> Web Component Only Icon
3. Check that no prop is being set if only icon is true and that the component renders as i it should-

**Screenshots**  
-

**Additional context**  
-


nathalielindqvist (commented):
When button size is "Extra small" (and only the button text is visible), should the control for "Icon Type" disappear the same way as the "Only Icon" and "Icon" controls do?

nathalielindqvist (approved):
👍

PR #820, opened 2023-02-08
feat(datetime): mode variant

EmelieLitwin:
**Describe pull-request**  
Dark mode variants for datetime component. Component need refactoring/redesign but this is better than nothing since component background is white and is going to look transparent on our light primary mode variant 

**Solving issue**  
Fixes: [AB#3188](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/3188)

**Suggested test steps**
- [ ] Storybook controls

### DIFF
--- a/tegel/src/components/datetime/datetime-theme-vars.scss
+++ b/tegel/src/components/datetime/datetime-theme-vars.scss
@@ -3,7 +3,7 @@
   --sdds-datetime-color: var(--sdds-grey-868);
   --sdds-datetime-background-primary: var(--sdds-grey-50);
   --sdds-datetime-background-secondary: var(--sdds-white);
-  --sdds-datetime-background: var( --sdds-datetime-background-primary)
+  --sdds-datetime-background: var( --sdds-datetime-background-primary);
   --sdds-datetime-icon: var(--sdds-grey-868);
   --sdds-datetime-border-bottom: var(--sdds-grey-400);
   --sdds-datetime-placeholder: var(--sdds-grey-500);
@@ -42,15 +42,11 @@
   // Prefix and Suffix
   --sdds-datetime-ps-color: var(--sdds-grey-600);
 }
-.sdds-mode-variant-primary {
-  --sdds-datetime-background: var(--sdds-datetime-background-primary)
-}
-.sdds-mode-variant-secondary {
-  --sdds-datetime-background: var(--sdds-datetime-background-secondary)
-}
 .sdds-theme-dark {
   --sdds-datetime-color: var(--sdds-grey-600);
-  --sdds-datetime-background: var(--sdds-grey-900);
+  --sdds-datetime-background-primary: var(--sdds-grey-900);
+  --sdds-datetime-background-secondary: var(--sdds-grey-868);
+  --sdds-datetime-background: var( --sdds-datetime-background-primary);
   --sdds-datetime-icon: var(--sdds-grey-50);
   --sdds-datetime-border-bottom: var(--sdds-grey-600);
   --sdds-datetime-placeholder: var(--sdds-grey-500);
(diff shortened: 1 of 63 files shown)


tegel/src/components/datetime/datetime-theme-vars.scss
```diff
@@ -1,14 +1,16 @@
 :root,
 .sdds-theme-light {
   --sdds-datetime-color: var(--sdds-grey-868);
-  --sdds-datetime-background: var(--sdds-white);
+  --sdds-datetime-background-primary: var(--sdds-grey-50);
+  --sdds-datetime-background-secondary: var(--sdds-white);
+  --sdds-datetime-background: var( --sdds-datetime-background-primary);
   --sdds-datetime-icon: var(--sdds-grey-868);
```
nathalielindqvist, line 7: The icon seems to be Grey958 in light mode in Figma
image

Icon also changes color to red when put in error state in the design, is there a reason as to why that is not the case in the component? :relaxed:
image

tegel/src/components/datetime/datetime.stories.tsx
```diff
@@ -110,6 +121,7 @@ export default {
     labelText: 'Label text',
     helper: true,
```
nathalielindqvist, line 122: There is no help text in the design, only when in error state. Should we still have a general helper text in the code? Has there been a discussion about it that is not reflected in Figma?

tegel/src/components/datetime/datetime-theme-vars.scss
```diff
@@ -1,14 +1,16 @@
...
   --sdds-datetime-border-bottom: var(--sdds-grey-400);
   --sdds-datetime-placeholder: var(--sdds-grey-500);
   --sdds-datetime-border-bottom-hover: var(--sdds-grey-600);
```
nathalielindqvist, line 11: Might not be part of this PR but the bottom border in success state seems to be Grey958 in light mode in Figma
image
image

tegel/src/components/datetime/datetime.stories.tsx
```diff
@@ -98,6 +98,17 @@ export default {
...
     },
+    modeVariant: {
+      name: 'Mode variant',
+      description: 'Mode variant of the component.',
+      control: {
+        type: 'radio',
+      },
+      options: ['Inherit from parent', 'Primary', 'Secondary'],
+      table: {
+        defaultValue: { summary: 'Inherit from parent' },
```
nathalielindqvist, line 109: Should we have default values for the rest of the controls as well?

tegel/src/components/datetime/datetime-theme-vars.scss
```diff
@@ -1,14 +1,16 @@
...
 
   // Disabled
-  --sdds-datetime-background-disabled: var(--sdds-white);
+  --sdds-datetime-background-disabled: var(--sdds-grey-50);
   --sdds-datetime-color-disabled: var(--sdds-grey-400);
   --sdds-datetime-placeholder-disabled: var(--sdds-grey-400);
   --sdds-datetime-label-disabled: var(--sdds-grey-400);
```
nathalielindqvist, line 16: Might not be part of this PR, but the label in disabled state seems to use the wrong color token, in Figma it's Grey400 in light mode and Grey700 in dark mode.
image
image

image
image

mJarsater (commented):
Except for the issues raised by Nathalie this one look good to me!

tegel/src/components/datetime/datetime.stories.tsx
EmelieLitwin: > There is no help text in the design, only when in error state. Should we still have a general helper text in the code? Has there been a discussion about it that is not reflected in Figma?

Good find. Without a little research on best practice I'm not sure. I definitely think we should keep the prop, but probably add it to Figma as well. At least for success state...

PR #821, opened 2023-02-09
fix: mode variant add null and inherit from parent to components

EmelieLitwin:
**Describe pull-request**  
Card, message, textfield and textarea was missing null option to mode variant and/or Inherit from parent as option in storybook

**How to test**  
Check the components and that mode variants working in storybook and looks ok in code

### DIFF
--- a/tegel/src/components/message/sdds-message.tsx
+++ b/tegel/src/components/message/sdds-message.tsx
@@ -10,7 +10,7 @@ export class SddsMessage {
   @Prop() header: string;
 
   /** Variant of the component, based on current mode. */
-  @Prop() modeVariant: 'primary' | 'secondary' = 'primary';
+  @Prop() modeVariant: 'primary' | 'secondary' | null = null;
 
   /** Type of message. */
   @Prop() type: 'information' | 'error' | 'warning' | 'success' = 'information';
@@ -43,7 +43,7 @@ export class SddsMessage {
           class={`
         message-wrapper ${this.type}
         ${this.minimal ? 'message-minimal' : ''}
-        sdds-mode-variant-${this.modeVariant}`}
+        ${this.modeVariant !== null ? `sdds-mode-variant-${this.modeVariant}`: ''}`}
         >
           {!this.noIcon && <sdds-icon name={this.getIconName()} size="20px"></sdds-icon>}
           <div class={`message-content`}>
--- a/tegel/src/components/message/sdds-message.tsx
+++ b/tegel/src/components/message/sdds-message.tsx
@@ -10,7 +10,7 @@ export class SddsMessage {
   @Prop() header: string;
 
   /** Variant of the component, based on current mode. */
-  @Prop() modeVariant: 'primary' | 'secondary' | null = null;
+  @Prop() modeVariant: 'primary' | 'secondary' = null;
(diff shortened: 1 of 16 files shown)


charlottelorensson:
Skärmavbild 2023-02-09 kl  16 48 09

Perhaps not the place (and perhaps won't be so visible) but just something I noticed, it looks like the web component has blue lines on the corners.

And it seems like the font color in the textfields are not aligned with Figma:

Skärmavbild 2023-02-09 kl  17 00 10


But the mode variants seem to be working fine! :)

charlottelorensson:
Forgot to mention that the first image was the message component* :)

tegel/src/components/textarea/textarea.tsx
```diff
@@ -46,7 +46,7 @@ export class Textarea {
   @Prop() maxLength: number;
 
   /** Mode variant of the textarea */
-  @Prop() modeVariant: 'primary' | 'secondary' = 'primary';
+  @Prop() modeVariant: 'primary' | 'secondary' | null = null;
```
mJarsater, line 49:
```suggestion
  @Prop() modeVariant: 'primary' | 'secondary' = null;
```

tegel/src/components/message/sdds-message.tsx
```diff
@@ -10,7 +10,7 @@ export class SddsMessage {
   @Prop() header: string;
 
   /** Variant of the component, based on current mode. */
-  @Prop() modeVariant: 'primary' | 'secondary' = 'primary';
+  @Prop() modeVariant: 'primary' | 'secondary' | null = null;
```
mJarsater, line 13:
```suggestion
  @Prop() modeVariant: 'primary' | 'secondary' = null;
```

mJarsater (changes requested):
Nice 🚀! I think we can remove the optional null value on the modeVariant prop. Then it won't show in the readme and Im guessing we don't really want people to have to care about this.

EmelieLitwin:
> Nice 🚀! I think we can remove the optional null value on the modeVariant prop. Then it won't show in the readme and Im guessing we don't really want people to have to care about this.

Good feedback! If that's something you think we should do, it should be done on all components!

mJarsater:
> > Nice 🚀! I think we can remove the optional null value on the modeVariant prop. Then it won't show in the readme and Im guessing we don't really want people to have to care about this.
> 
> Good feedback! If that's something you think we should do, it should be done on all components!

Yeah that makes sense to me. Then all components with a modeVariant would have:
`@Prop() modeVariant: 'primary' | 'secondary' = null;`

EmelieLitwin:
> Skärmavbild 2023-02-09 kl 16 48 09
> 
> Perhaps not the place (and perhaps won't be so visible) but just something I noticed, it looks like the web component has blue lines on the corners.
> 
> And it seems like the font color in the textfields are not aligned with Figma:
> 
> Skärmavbild 2023-02-09 kl 17 00 10
> 
> But the mode variants seem to be working fine! :)

What is he problem with the textfield? In dark mode the inside label is grey-400 and placeholder grey-600, looks like it's the same in figma?

mJarsater (approved):
LGTM!

PR #795, opened 2023-01-20
feat(header, side-menu): webcomponents

alexandereneroth:
Experimental v2 header and side menu web components.

Looking for feedback on the API. 

Some notes:
- I've exposed the elements within the menus as individual web components for flexibility.
- The drawer open state is handled by the user to have a single source of truth.

To view, open the preview site below 👇

**Screenshots**

Screenshot 2023-01-20 at 16 33 10

**TODO**

- [x] Launcher menu
- [x] Drawer sub-menus
- [x] Header dropdowns
- [x] Simplify side menu user API
- [x] Fix collapsed side-submenu overflow bug
- [x] Always show collapse button in view when side menu has scroll
- [x] Handle basic text overflow in header
- [x] Align user menu in header right
- [x] Rename components
- [x] Selected states, hover, focus
- [x] Side menu width
- [x] Property descriptions
- [x] Event names
- [x] Clean up stories
- [x] Clean up CSS variables
- [x] Rebase
- [x] SonarCloud

PR #841, opened 2023-02-16
feat(popover-canvas): make unopinionated

alexandereneroth:
Makes the popover-canvas more useful by making less assumptions in terms of what content the user will put inside it, and in what context it will be used.

This PR:
- Removes the padding inside popover canvas
- Removes min-height and min-width of popover canvas
- Makes its opened state controllable from the outside with a "show" prop, the previous "show" prop which only controlled its initial state has been renamed "initialShow"
- Enables overriding/adding modifiers to the internal popper function

**Solving issue**  
Does not directly solve an issue by itself, but will be part of #795

**How to test**  
Look at the storybook in the deploy preview link below.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


tegel/src/components/popover-canvas/popover-canvas.tsx
```diff
@@ -8,7 +8,7 @@ import type { Placement, Instance } from '@popperjs/core';
   shadow: true,
 })
 export class PopoverCanvas {
-  @Element() popoverCanvasElement!: HTMLElement;
+  @Element() popoverCanvasElement!: HTMLSddsPopoverCanvasElement;
```
mJarsater, line 11: Nice!

tegel/src/components/popover-canvas/popover-canvas.tsx
```diff
@@ -17,7 +17,12 @@ export class PopoverCanvas {
   @Prop() referenceEl: HTMLElement;
 
   /** Decides if the Popover Canvas should be visible from the start */
-  @Prop() show: boolean = false;
+  @Prop() initialShow: boolean = false;
+
+  /** Controls wether the popover is shown or not. If this is set hiding and showing
+   * will be decided by this prop and will need to be controlled from the outside.
+   */
+  @Prop() show: boolean = null;
```
mJarsater, line 25: Why do we need both of these? Wouldn't just show work, feels like it would be more aligned to other components.

tegel/src/components/popover-canvas/popover-canvas.tsx
```diff
@@ -28,24 +33,40 @@ export class PopoverCanvas {
   /** Sets the offset distance */
   @Prop() offsetDistance: number = 8;
 
+  @Prop() modifiers: Object[] = [];
```
mJarsater, line 36: Should we add a comment to explain this one? And if not, mark it as @internal so it doesn't show in the readme.

tegel/src/components/popover-canvas/popover-canvas.tsx
```diff
@@ -54,43 +75,68 @@ export class PopoverCanvas {
...
 
-    const hideCanvas = () => {
-      this.show = false;
-    };
+  @Listen('click', { target: 'window' })
+  handleOutsideClick() {
+    if (this.show === null && this.isShown) {
+      this.isShown = false;
+    }
+  }
```
mJarsater, line 107: Nothing we need to fix now, but how does this work with a11y? If the user tabs out of the popover, they need to click something for it to close?

tegel/src/components/popover-canvas/popover-canvas.tsx
```diff
@@ -54,43 +75,68 @@ export class PopoverCanvas {
...
 
-      if (this.show) {
-        hideCanvas();
+      if (referenceEl) {
+        this.initialize(referenceEl);
       } else {
-        showCanvas();
+        console.error(
+          `Could not initialize popover-canvas: element with selector '${this.selector}' not found.`,
+        );
```
mJarsater, line 118: Nice! 🦄

mJarsater (approved):
LGTM! 🧁

PR #845, opened 2023-02-17
fix: null not valid value for mode variant prop

EmelieLitwin:
**Describe pull-request**  
Some component still had null as valid input. Also fixed inline-tab sb.


mJarsater (approved):
👍

Issue #767, opened 2023-01-04
[Bug report]: Textfield component not emitting "focus" and "blur" events anymore


bogdan-todireanu:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have followed the installation guide.

### Package

@scania/theme-light

### Package versions

@scania/components@4.10.4

### Browser

Chrome

### Framework

Angular

### Version

Angular 15

### Reproduction steps

1. Add component <sdds-textfield> (@scania/components@4.10.4) in an Angular 15 project
2. Listen to "blur" and "focus" events
3. The above mentioned events won't be triggered

The last version of @scania/components that still triggered the events is 4.9.2. 

### Code example

`<sdds-textfield
    type="text"
    size="md"
    state="{{ !location && !disabled && !loading ? 'error' : '' }}"
    (input)="onInput($event)"
    (blur)="onBlur()" // onBlur not getting triggered
    (focus)="onFocus()" // onFocus not getting triggered
    value="{{ location }}"
    [disabled]="disabled"
    class="sdds-on-white-bg"
  >
  </sdds-textfield>`

### Screenshots

Add/Copy screenshot here

### Expected behaviour

Component Textfield should be emit events when the actual input is focused or blurred 

### Console errors

Add warning/error message

### Contact information

bogdan.todireanu@scania.com

Issue #759, opened 2023-01-02
[Feature Request]: Trigger onModalCloseEvent when hitting the X button on modals.


toliv3:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md)
- [X] I have checked the lastest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Trigger an event (suggestion: onModalCloseEvent) when hitting the X button so we can have programmatic control over what happens in the modal.

### Background

The current modal lacks programmatic control. We need to have control of what happens to be able to migrate existing functionality to sdds/tegel.

### Contact information

thiago.oliveira@scania.com

mJarsater:
Fixed in: https://github.com/scania-digital-design-system/sdds/releases/tag/%40scania%2Fcomponents%404.10.7

The events is called `sddsClose`

PR #850, opened 2023-02-20
feat(events - textfield/textarea/modal): updated events

mJarsater:
**Describe pull-request**  
Updated the focus/blur/input/change event for the textfield/textarea. Added a sddsClose event for the modal.

**Solving issue**  
Fixes: #767 #759

**How to test**  
1. Check out the branch
2. Run npm run storybook for components
3. Check the readme

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/components/src/components/textarea/readme.md
+++ b/components/src/components/textarea/readme.md
@@ -27,10 +27,10 @@
 
 | Event          | Description                   | Type                      |
 | -------------- | ----------------------------- | ------------------------- |
+| `customBlur`   | Blur event for the textarea   | `CustomEvent<FocusEvent>` |
 | `customChange` | Change event for the textarea | `CustomEvent<any>`        |
-| `sddsBlur`     | Blur event for the textarea   | `CustomEvent<FocusEvent>` |
-| `sddsFocus`    | Focus event for the textarea  | `CustomEvent<FocusEvent>` |
-| `sddsInput`    | Input event for the textarea  | `CustomEvent<InputEvent>` |
+| `customFocus`  | Focus event for the textarea  | `CustomEvent<FocusEvent>` |
+| `customInput`  | Input event for the textarea  | `CustomEvent<InputEvent>` |
 
 
 ----------------------------------------------
--- a/components/src/components/textarea/readme.md
+++ b/components/src/components/textarea/readme.md
@@ -25,9 +25,12 @@
 
 ## Events
 
-| Event          | Description                   | Type               |
-| -------------- | ----------------------------- | ------------------ |
-| `customChange` | Change event for the textarea | `CustomEvent<any>` |
+| Event        | Description                   | Type                      |
+| ------------ | ----------------------------- | ------------------------- |
+| `sddsBlur`   | Blur event for the textarea   | `CustomEvent<FocusEvent>` |
+| `sddsChange` | Change event for the textarea | `CustomEvent<any>`        |
+| `sddsFocus`  | Focus event for the textarea  | `CustomEvent<FocusEvent>` |
+| `sddsInput`  | Input event for the textarea  | `CustomEvent<InputEvent>` |
 
 
 ----------------------------------------------
--- a/components/src/components/textarea/readme.md
+++ b/components/src/components/textarea/readme.md
@@ -25,12 +25,12 @@
 
 ## Events
 
-| Event        | Description                   | Type                      |
-| ------------ | ----------------------------- | ------------------------- |
-| `sddsBlur`   | Blur event for the textarea   | `CustomEvent<FocusEvent>` |
-| `sddsChange` | Change event for the textarea | `CustomEvent<any>`        |
-| `sddsFocus`  | Focus event for the textarea  | `CustomEvent<FocusEvent>` |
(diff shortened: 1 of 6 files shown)


components/src/components/textarea/textarea.tsx
```diff
@@ -53,25 +53,62 @@ export class Textarea {
 
   /** Change event for the textarea */
   @Event({
+    eventName: 'sddsChange',
     composed: true,
     bubbles: true,
-    cancelable: true,
+    cancelable: false,
   })
-  customChange: EventEmitter;
```
alexandereneroth, line 60: Changing the event name here would be a breaking change, maybe cancelable shouldn't be changed as well for the same reason, although i admit i'm not sure what its for here.

If we keep custom change here, then i think the other events could also follow the "customSomething" naming pattern

components/src/components/textarea/textarea.tsx
```diff
@@ -53,25 +53,62 @@ export class Textarea {
...
+    cancelable: false,
+  })
+  sddsFocus: EventEmitter<FocusEvent>;
+
+  /* Set the input as focus when clicking the whole textfield with suffix/prefix */
+  handleFocus(event): void {
+    if (!this.readonly) {
+      this.textEl.focus();
+      this.focusInput = true;
+      this.sddsFocus.emit(event);
```
alexandereneroth, line 110: I don't think this method is needed since the textarea already has a onFocus on it. Id move the focusInput=true and emit focus inside there instead.

alexandereneroth (changes requested):

components/src/components/textarea/readme.md
```diff
@@ -25,9 +25,12 @@
 
 ## Events
 
-| Event          | Description                   | Type               |
-| -------------- | ----------------------------- | ------------------ |
-| `customChange` | Change event for the textarea | `CustomEvent<any>` |
+| Event          | Description                   | Type                      |
+| -------------- | ----------------------------- | ------------------------- |
+| `customChange` | Change event for the textarea | `CustomEvent<any>`        |
+| `sddsBlur`     | Blur event for the textarea   | `CustomEvent<FocusEvent>` |
```
alexandereneroth, line 31: IMO its better with a bad name as long as its consistent, so I'd use customBlur, customFocus, etc here.
Then we improve our users chances of learning the API, and improve readability

components/src/components/textarea/textarea.tsx
```diff
@@ -56,25 +56,62 @@ export class Textarea {
 
   /** Change event for the textarea */
   @Event({
+    eventName: 'customChange',
     composed: true,
     bubbles: true,
-    cancelable: true,
+    cancelable: false,
```
alexandereneroth, line 62: Are you sure this change is not breaking the API in some way?

components/src/components/textfield/textfield.tsx
```diff
@@ -110,24 +152,13 @@ export class Textfield {
           <slot name="sdds-label" />
         </div>
 
-        <div onClick={() => this.handleFocusClick()} class="sdds-textfield-container">
+        <div onClick={() => this.textInput.focus()} class="sdds-textfield-container">
```
alexandereneroth, line 155: In my limited testing, this focus() call did not seem to be needed.

components/src/components/textfield/textfield.tsx
```diff
@@ -50,20 +50,62 @@ export class Textfield {
...
+    eventName: 'sddsFocus',
+    composed: true,
+    bubbles: true,
+    cancelable: false,
+  })
+  sddsFocus: EventEmitter<FocusEvent>;
+
+  /** Set the input as focus when clicking the whole textfield with suffix/prefix */
+  handleFocus(event): void {
+    console.log('hej');
```
alexandereneroth, line 90:
```suggestion
```

alexandereneroth (changes requested):

components/src/components/textarea/textarea.tsx
mJarsater: Yes, since we are not checking for any cancelation. But since it's not really visible I can remove it.

components/src/components/textarea/readme.md
mJarsater: Agree! Will rename the other events

components/src/components/textfield/textfield.tsx
```diff
@@ -50,20 +50,62 @@ export class Textfield {
...
+    composed: true,
+    bubbles: true,
+    cancelable: false,
+  })
+  sddsInput: EventEmitter<InputEvent>;
 
   // Data input event in value prop
-  handleInput(e): void {
-    this.value = e.target.value;
+  handleInput(event): void {
```
timrombergjakobsson, line 74: Just a question, but don't we want to update the value before emitting? To mitigate that the value could be out of sync? Just thinking out loud.
```suggestion
  this.value = event.target.value;
  this.sddsInput.emit(event);
```

mJarsater: Absolutely!

alexandereneroth (approved):
🎖️

Issue #525, opened 2022-10-21
[Bug report]: sdds-modal should allow scroll only on the body


toliv3:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have made sure that the issue comes from SDDS packages by removing SDDS from my project as a test.
- [X] I have followed the installation guide.

### Package

@scania/tegel-old

### Package versions

@scania/components@latest

### Browser

Chrome

### Framework

Angular

### Reproduction steps

1. Create a modal.
2. Add a sdds-text-field inside the modal
3. Add more content to provoke scrolling
4. Check that when scrolling the sdds-text-field will be above the modal header instead of below.

### Code example

<div>....

### Screenshots

![image](https://user-images.githubusercontent.com/111114201/197161368-c82e3ff0-f7b5-4861-826c-9d40fa23cfb1.png)

![image](https://user-images.githubusercontent.com/111114201/197161314-d230f3ee-3995-4b76-870f-dfaf09f83012.png)

### Expected behaviour

Scrolling on sdds-modal should only affect the body of the modal, currently as it scrolls the whole component, when scrolling with a defined header (sdds-modal-headline), other components like the sdds-text-field will scroll above the headline.

### Console errors

Add warning/error message

### Contact information

thiago.oliveira@scania.com

[AB#2444](https://dev.azure.com/scaniadigitaldesignsystem/3d9eb9c6-0045-473d-bc1d-e842dd779200/_workitems/edit/2444)

PR #852, opened 2023-02-21
fix(modal): fix overflow/z-index issue within modal

mJarsater:
**Describe pull-request**  
Added a high z-index for the modal header in order for all elements that needs to scroll goes behind it.

**Solving issue**  
Fixes: #525

**How to test**  
1. Go to checkout branch
2. Run storybook for components
3. Add som elements to the scroll body and make sure the header scroll "above" all elements.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/components/src/components/modal/_modal-core.scss
+++ b/components/src/components/modal/_modal-core.scss
@@ -160,6 +160,7 @@ $modals: (
   position: sticky;
   top: 0;
   background-color: var(--sdds-modal-bg);
+  z-index: 1000;
 
   &::before {
     content: ' ';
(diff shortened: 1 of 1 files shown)


components/src/components/modal/_modal-core.scss
```diff
@@ -160,6 +160,7 @@ $modals: (
   position: sticky;
   top: 0;
   background-color: var(--sdds-modal-bg);
+  z-index: 1000;
```
alexandereneroth, line 163:
```suggestion
  z-index: 1;
```
I think it should be as low as possible, since we just want to fix that elements with `position:relative` overlaps the header.
If we keep it low we still allow for tooltips, popovers, and other things to overlay the header.

timrombergjakobsson (approved):

PR #873, opened 2023-02-28
Feat/dropdown with shadow dom

mJarsater:
**Describe pull-request**  
Version 2 of the dropdown. This component has all the functionality for multiselect/filter in one component. Also adds a hidden input element outside of the shadowDOM to be able to use the dropdown within a form. This includes two new element: sdds-dropdown-v2 and sdds-dropdown-option-v2. First one is a wrapper element that holds the dropdown button and most of the logic. The second is child element that should represent all options in the dropdown. As oppose to previous versions of the component we are using the default slot on the sdds-dropdown-option to set a label. This is closer to the browser implementation of select/option.

## Note!
This component does not accept a "data" prop. This is due to recommendation for webcomponent to not pass arrays/objects as props/attributes. Instead the user would have to map over and render how many sdds-dropdown-options they my need.

**Solving issue**  
Fixes: AB#3174

**How to test**  
2. Check in Components -> Dropdown-v2
3. Test the controls and check the API.  

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


tegel/src/components/dropdown-v2/dropdown-option-v2/sdds-dropdown-option-v2.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  connectedCallback = () => {
+    this.parentElement = this.host.parentElement as HTMLSddsDropdownV2Element;
+    this.modeVariant = this.parentElement.modeVariant;
+    this.multiselect = this.parentElement.multiselect;
+    this.size = this.parentElement.size;
+    this.label = this.host.textContent.trim();
+  };
+
+  handleSingleselect = () => {
```
timrombergjakobsson, line 46: Small thing but maybe should be `handleSingleSelect`?

mJarsater: Yes!

alexandereneroth (changes requested):

alexandereneroth:
I'll leave some comments separately so they can be separate threads...

I think it would be good if inputs had a default width for cases where they should not fill the entire container. In Figma it seems to be 208px, also it is documented here: https://tegel.scania.com/components/textfields#:~:text=text%20fields%20is-,208px,-but%20this%20can

MUI has a fullWidth prop instead, for example: https://mui.com/material-ui/react-text-field/#full-width

alexandereneroth:
Some deviations from Figma: https://www.figma.com/file/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=9754%3A22916&t=PKbKkCMJzUjnw0uH-0 

- The dropdown should have a hover state where the underline gets darker as you hover
- The inner label should be grey 700 in light mode, currently it's 958
- When the input has an inner label; the label and value are not aligned in the vertical center of the input
- The dropdown item height should be 48px. Unclear what the small size should have though (40px vs 48px @charlottelorensson )
- The dropdown input should not get its focused appearance on click, only when it is possible to type in it (tip ::focus-visible)
- The whole component should get greyed out when disabled, not just the value
- When small size, and label inside are set, the value gets misaligned

alexandereneroth:
I'd align the terminology "child", "option", and "data" to avoid confusion, since they are pretty much referring to the same thing.
In Storybook there is "Child type" and "Default options" for example. There is also a property "data" on the component. 
I suggest using "option" for this since, the child elements are called option.

"Child type" could be "Set options with", for example. And the "data" prop, could be called "options".

tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx
```diff
@@ -0,0 +1,351 @@
...
+  @Prop() filter: boolean = false;
+
+  /** Text that is displayed if filter is used and there are no options that matches the search. */
+  @Prop() noResultText: string = 'No result';
+
+  /** Default value selected in the dropdown. */
+  @Prop() defaultValue: string;
+
+  /** Populate the dropdown via a JSON array */
+  @Prop() data: string;
```
alexandereneroth, line 54: I think this could be an array of objects

alexandereneroth (changes requested):

tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+  @State() open: boolean = false;
+
+  @State() value: Array<string>;
+
+  @State() valueLabels: Array<string>;
```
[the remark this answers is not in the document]
mJarsater: Agree, separated them to keep value separate as it is the value being passed to the hidden input. But I agree, will fix!

mJarsater:
> Some deviations from Figma: https://www.figma.com/file/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=9754%3A22916&t=PKbKkCMJzUjnw0uH-0
> 
> * [ ]  When the input has an inner label; the label and value are not aligned in the vertical center of the input
@charlottelorensson )
> * [ ]  The dropdown input should not get its focused appearance on click, only when it is possible to type in it (tip ::focus-visible)

I don't think it suppose to be aligned vertically when there is a inner label. In the design it has 16px to bottom and 24px to top. 

Do you mean the blue underline? Maybe we can get some input from @charlottelorensson on that one? 😄

mJarsater:
> I'd align the terminology "child", "option", and "data" to avoid confusion, since they are pretty much referring to the same thing. In Storybook there is "Child type" and "Default options" for example. There is also a property "data" on the component. I suggest using "option" for this since, the child elements are called option.
> 
> "Child type" could be "Set options with", for example. And the "data" prop, could be called "options".

Great feedback!

charlottelorensson:
> > Some deviations from Figma: https://www.figma.com/file/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=9754%3A22916&t=PKbKkCMJzUjnw0uH-0
> > 
> > * [ ]   When the input has an inner label; the label and value are not aligned in the vertical center of the input
> >   @charlottelorensson )
> > * [x]   The dropdown input should not get its focused appearance on click, only when it is possible to type in it (tip ::focus-visible)
> 
> I don't think it suppose to be aligned vertically when there is a inner label. In the design it has 16px to bottom and 24px to top.
> 
> Do you mean the blue underline? Maybe we can get some input from @charlottelorensson on that one? 😄

So sorry I've missed this. Yes, the small size has 40px but I'm a bit unsure what you mean with the alignment, let's have a chat about it tomorrow!

alexandereneroth:
> deploy-preview-873--tegel-storybook.netlify.app

@mJarsater @charlottelorensson 

What I meant is that its not centered vertically:
Screenshot 2023-03-09 at 13 20 23

But now when i look in Figma is not centered vertically there either... So NVM, my eyes must have betrayed me.

tegel/src/components/dropdown-v2/dropdown-option-v2/sdds-dropdown-option-v2.tsx
```diff
@@ -0,0 +1,176 @@
...
+    cancelable: false,
+  })
+  sddsBlur: EventEmitter<FocusEvent>;
+
+  connectedCallback = () => {
+    this.parentElement =
+      this.host.parentElement.tagName === 'SDDS-DROPDOWN-V2'
+        ? (this.host.parentElement as HTMLSddsDropdownV2Element)
+        : ((this.host.getRootNode() as ShadowRoot).host as HTMLSddsDropdownV2Element);
+    this.modeVariant = this.parentElement.modeVariant;
```
alexandereneroth, line 81: Wouldn't this be redundant, as css variables leak in through shadow dom. So if an element has a mode variant, all its children would get that mode variant anyway?

mJarsater: You are absolutely correct! Nice catch 🎣

mJarsater:
> > deploy-preview-873--tegel-storybook.netlify.app
> 
> @mJarsater @charlottelorensson
> 
> What I meant is that the label and value together is not centered vertically: Screenshot 2023-03-09 at 13 20 23
> 
> But now when i look in Figma is not centered vertically there either... So NVM, my eyes must have betrayed me. I'm being too nitpicky anyway (:

You are the perfect amount of nitpicky!

tegel/src/components/dropdown-v2/dropdown-option-v2/sdds-dropdown-option-v2.tsx
```diff
@@ -0,0 +1,173 @@
...
+  @Method()
+  async deselect() {
+    this.selected = false;
+  }
+
+  /** Method to select the dropdown option. */
+  @Method()
+  async select() {
+    this.selected = true;
+  }
```
alexandereneroth, line 42: What about a single `setSelected` instead? 
Just feels better to me with a single method instead of two.
Remember that any methods we expose are setting a convention for other components to use.

Alternatively a @Prop with a watcher perhaps?

tegel/src/components/dropdown-v2/sdds-dropdown-v2.scss
```diff
@@ -0,0 +1,345 @@
+@import '../../mixins/z-index';
```
timrombergjakobsson, line 1: Just a thought, but as this file is so long, would it be beneficial to split it up into some smaller files? For more readability, like all styles applied to ```.dropdown-select```

tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx
```diff
@@ -0,0 +1,405 @@
...
+    if (this.defaultValue) {
+      this.setDefaultOption();
+    }
+    if (this.openDirection === 'auto') {
+      this.getOpenDirection();
+    }
+  }
+
+  setDefaultOption = () => {
+    this.children = this.children.map((element: HTMLSddsDropdownOptionV2Element) => {
```
timrombergjakobsson, line 196: Maybe use ```forEach``` here instead of ```map```? As we are only iterating over the array right?

mJarsater: Maybe! I was thinking that we are altering the values in the array (when we do element.select()), so I want to return a fresh array to make sure we are always working with the most up to date children. Maybe a forEach instead won't affect that though. Im just very used to never alter an array, and if so do it in a new one instead.

tegel/src/components/dropdown-v2/sdds-dropdown-v2.scss
mJarsater: I like this idea. Were you thinking we would have one base file that just imports the other ones? So we could have dropdown-filter.scss, dropdown-multiselect.scss and then dropdown.sccs which just imports the other two and is the one referenced in the stencil component?

timrombergjakobsson: > I like this idea. Were you thinking we would have one base file that just imports the other ones? So we could have dropdown-filter.scss, dropdown-multiselect.scss and then dropdown.sccs which just imports the other two and is the one referenced in the stencil component?

Yes exactly, I think it would improve readability! I personally prefer more files to one big file.

tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx
timrombergjakobsson: Yes I see what you mean, I think I was reading it as though we are not altering the values really.

tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx
```diff
@@ -0,0 +1,405 @@
...
+  };
+
+  render() {
+    renderHiddenInput(
+      this.host,
+      this.name,
+      this.selection?.map((item) => item.value).toString(),
+      this.disabled,
+    );
+    return (
```
timrombergjakobsson, line 264: Would it be helpful to split this part of the code up into smaller pieces also? Maybe one way could be
```suggestion
    renderLabelOutside() {
      if (!this.label || this.labelPosition !== 'outside') {
        return null;
      }
      return (
        <div class={`label-outside ${this.disabled ? 'disabled' : ''}`}>
          {this.label}
        </div>
      );
}
```

And then the whole part with dropdown select could maybe be a function:

```suggestion
    renderDropdownSelect() {
        const filter = this.filter;
        const selection = this.selection;
        const labelPosition = this.labelPosition;
        const disabled = this.disabled;
        const size = this.size;
      
        if (filter) {
          return (
            <div class={`filter ${this.filterFocus ? 'focus' : ''} ${disabled ? 'disabled' : ''}`}>
              <div class="value-wrapper">
```
etc. 

And then be used
```suggestion
  return (
    <Host class={`${this.modeVariant ? `sdds-mode-variant-${this.modeVariant}` : ''}`}>
      {this.renderLabelOutside()}
      <div class={`dropdown-select ${this.size} ${this.disabled ? 'disabled' : ''}`}>
        {this.renderDropdownSelect()}
      </div>
      {/* DROPDOWN LIST */}
      <div
        ref={(element) => (this.dropdown
  ```

Comment (mJarsater) [reply tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx]: I tried this out locally, but I find it very hard to navigate the render function. when there are other functions also returning HTML. Maybe something that can be discussed in a checkin?

Comment (timrombergjakobsson) [reply tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx]: I totally understand, maybe something to look at together?

Comment (mJarsater) [reply tegel/src/components/dropdown-v2/sdds-dropdown-v2.tsx]: Absolutely!

File: tegel/src/components/dropdown-v2/dropdown-v2.tsx [hunk 9]
```diff
@@ -0,0 +1,406 @@
...
+
+  /** Text that is displayed if filter is used and there are no options that matches the search. */
+  @Prop() noResultText: string = 'No result';
+
+  /** Default value selected in the dropdown. */
+  @Prop() defaultValue: string;
+
+  // This has to accepts a string to work for vanilla HTML.
+  /** Populate the dropdown via a JSON array */
+  @Prop() options: Array<{ value: string; label: string; disabled: boolean }> | string;
```
alexandereneroth, line 55: According to Googles "custom elements best practices" rich data(objects & arrays) should not be accepted as attributes, since they have to be stringified for this to be possible (https://web.dev/custom-elements-best-practices/#:~:text=Aim%20to%20only%20accept%20rich%20data%20(objects%2C%20arrays)%20as%20properties.) 

Stencil also doesn't stringify rich data automatically for the same reason: (https://stenciljs.com/docs/properties#object-props:~:text=It%20is%20not%20possible%20to%20set%20Object%20props%20via%20an%20HTML%20attribute%20like%20so%3A)

You can pass rich data to an element in vanilla with javascript like this:
```javascript
document.getElementById('my-dropdown-1').options = [
  {
    value: 1, 
    label: 'ett'
  },
  {
    value: 2, 
    label: 'två'
  },
  {
    value: 3, 
    label: 'tre'
   }
]
```

alexandereneroth: I realize this is something we support in some other components as well, for example datatable.

alexandereneroth (changes requested):

tegel/src/components/dropdown-v2/dropdown-v2.tsx
mJarsater: Either that, or we need to change our examples and stories to conform with best practices.

tegel/src/utils/utils.ts
```diff
@@ -19,3 +19,21 @@ export const formatHtmlPreview = (htmlStr) =>
 export function demoFormat(first: string, middle: string, last: string): string {
   return (first || '') + (middle ? ` ${middle}` : '') + (last ? ` ${last}` : '');
 }
+
+export const renderHiddenInput = (
```
timrombergjakobsson, line 23: Question, what is the responsibility of this function?

mJarsater: Of the renderHiddenInput? 

Is appends a input with type hidden to the element (in this case sdds-dropdown-v2). This enables the dropdown to be used within a form and still use the shadowDOM. If we have a shadowDOM inside a input element (like sdds-checkbox) it won't be able to get picked up by a form, so this is a workaround. 

The dropdown is the first component that we are trying this approach on, but eventually I would like us to use the same approach on all input element.

timrombergjakobsson: > Of the renderHiddenInput?
> 
> Is appends a input with type hidden to the element (in this case sdds-dropdown-v2). This enables the dropdown to be used within a form and still use the shadowDOM. If we have a shadowDOM inside a input element (like sdds-checkbox) it won't be able to get picked up by a form, so this is a workaround.
> 
> The dropdown is the first component that we are trying this approach on, but eventually I would like us to use the same approach on all input element.

Ok great! Thanks for the explanation!

tegel/src/utils/utils.ts
```diff
@@ -19,3 +19,21 @@ export const formatHtmlPreview = (htmlStr) =>
...
+  element: HTMLElement,
+  name: string,
+  value: string | undefined | null,
+  disabled: boolean,
+) => {
+  /* Maybe remove this and never have as input within the shadowDOM?. */
+  let input = element.querySelector('input') as HTMLInputElement | null;
```
timrombergjakobsson, line 30: Is null "enough" or also include `undefined`? So `let input = element.querySelector('input') as HTMLInputElement | null | undefined;`?

mJarsater: I don't think the querySelector will ever return undefined?

alexandereneroth: True, it only returns an Element or null https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#return_value

tegel/src/utils/utils.ts
```diff
@@ -19,3 +19,89 @@ export const formatHtmlPreview = (htmlStr) =>
...
+    childElement.innerHTML = slot;
+    childElement.id = id;
+    parentElement.appendChild(childElement);
+  }
+};
+
+/**
+ * Find the nexxt focusable item index in a list of focusable elements.
+ *
+ * @param options List of focusable elements, disabled items needs to have a disabled attribute.
```
mJarsater, line 78: Maybe rename this to items?

tegel/src/components/dropdown-v2/dropdown-option-v2/dropdown-option-v2.tsx
```diff
@@ -0,0 +1,171 @@
+import { Component, Host, h, Prop, State, Element, Event } from '@stencil/core';
+import { EventEmitter, HostElement, Method } from '@stencil/core/internal';
+import { SddsCheckboxCustomEvent } from '../../../components';
+
+@Component({
+  tag: 'sdds-dropdown-option-v2',
+  styleUrl: 'dropdown-option-v2.scss',
+  shadow: {
```
timrombergjakobsson, line 8: What is this one for?

timrombergjakobsson: Ok nice one!

tegel/src/components/dropdown-v2/dropdown-v2.stories.tsx
```diff
@@ -0,0 +1,303 @@
...
+      options: ['No default', 'Option 1', 'Option 2', 'Option 3'],
+      if: { arg: 'multiselect', eq: false },
+    },
+    multiDefaultOption: {
+      name: 'Default options',
+      description: 'Sets a pre-selected option.',
+      control: {
+        type: 'check',
+      },
+      options: ['Option 1', 'Option 2', 'Option 3', 'Option 3'],
```
nathalielindqvist, line 135:
```suggestion
      options: ['Option 1', 'Option 2', 'Option 3', 'Option 4'],
```

tegel/src/components/dropdown-v2/dropdown-v2.stories.tsx
```diff
@@ -0,0 +1,303 @@
...
+    defaultOption: {
+      name: 'Default options',
+      description: 'Sets a pre-selected option.',
+      control: {
+        type: 'radio',
+      },
+      options: ['No default', 'Option 1', 'Option 2', 'Option 3'],
+      if: { arg: 'multiselect', eq: false },
+    },
+    multiDefaultOption: {
```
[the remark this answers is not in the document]
mJarsater: Pretty sure this is a storybook "bug"

tegel/src/components/dropdown-v2/dropdown-option-v2/dropdown-option-vars.scss
```diff
@@ -0,0 +1,41 @@
...
+  --sdds-dropdown-option-background: var(--sdds-grey-50);
+  --sdds-dropdown-option-background-hover: var(--sdds-grey-100);
+  --sdds-dropdown-option-background-selected: var(--sdds-grey-300);
+  --sdds-dropdown-option-border: var(--sdds-grey-300);
+  --sdds-dropdown-label-color: var(--sdds-grey-958);
+  --sdds-dropdown-helper-color: var(--sdds-grey-700);
+  --sdds-dropdown-placeholder-color: var(--sdds-grey-700);
+
+  .sdds-mode-variant-primary {
+    --sdds-dropdown-option-background: var(--sdds-grey-50);
```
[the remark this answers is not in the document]
mJarsater: This is due to us setting the background/color in our globals.scss. It shouldn't be a problem since the issue will only appear if they are using it on the wrong background

nathalielindqvist (commented):
Seems like the alignment could be a bit off between the checkbox and the label when multiselect is active?
image

mJarsater:
> Seems like the alignment could be a bit off between the checkbox and the label when multiselect is active? image

Nice! Fixed it now and also moved the dropdown to a "Beta Components" section

timrombergjakobsson:
> Seems like the alignment could be a bit off between the checkbox and the label when multiselect is active? image

Yeah saw it too, most likely to the flex-start property on the checkbox?

nathalielindqvist (approved):
That beta looking good! 🔥

PR #902, opened 2023-03-03
fix(popover): update default values in Storybook

nathalielindqvist:
**Describe pull-request**  
Updated default values and descriptions in both popover canvas and popover menu components.

**How to test**  
1. Go to Storybook link below
2. Check in Popover Canvas -> Default
3. Check in Popover Menu -> Default and With Icons
4. Read control descriptions

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/tegel/src/components/popover-canvas/popover-canvas.stories.tsx
+++ b/tegel/src/components/popover-canvas/popover-canvas.stories.tsx
@@ -29,7 +29,9 @@ export default {
         'Right end',
         'Auto',
       ],
-      defaultValue: { summary: 'Bottom' },
+      table: {
+        defaultValue: { summary: 'Bottom' },
+      },
     },
   },
   args: {
--- a/tegel/src/components/popover-canvas/popover-canvas.stories.tsx
+++ b/tegel/src/components/popover-canvas/popover-canvas.stories.tsx
@@ -30,12 +30,12 @@ export default {
         'Auto',
       ],
       table: {
-        defaultValue: { summary: 'Bottom' },
+        defaultValue: { summary: 'Auto' },
       },
     },
   },
   args: {
-    canvasPosition: 'Bottom',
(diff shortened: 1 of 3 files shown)


tegel/src/components/popover-canvas/popover-canvas.stories.tsx
```diff
@@ -30,7 +29,9 @@ export default {
         'Right end',
         'Auto',
       ],
-      defaultValue: { summary: 'Auto' },
+      table: {
+        defaultValue: { summary: 'Bottom' },
```
mJarsater, line 33: I think this should be changed to 'auto', that's the default value that is set in code. Should also update the args object so it's set as Auto there.

tegel/src/components/popover-menu/popover-menu-icons.stories.tsx
```diff
@@ -40,10 +40,13 @@ export default {
...
         'Right end',
       ],
+      table: {
+        defaultValue: { summary: 'Bottom' },
+      },
     },
   },
   args: {
-    menuPosition: 'Bottom start',
+    menuPosition: 'Bottom',
```
mJarsater, line 49: I think this should be changed to 'auto', that's the default value that is set in code. Should also update the args object so it's set as Auto there.

tegel/src/components/popover-menu/popover-menu.stories.tsx
```diff
@@ -40,10 +40,13 @@ export default {
         'Right start',
         'Right end',
       ],
+      table: {
+        defaultValue: { summary: 'Bottom' },
```
mJarsater, line 44: I think this should be changed to 'auto', that's the default value that is set in code. Should also update the args object so it's set as Auto there.

PR #912, opened 2023-03-07
fix(chip): removed class from sdds-icon

mJarsater:
**Describe pull-request**  
This is to increase the interoperability with react. Web component should not need to have classes on them. From what I could find this was the only place where we were adding a class to a <sdds-icon> komponent.

**Solving issue**  
Fixes: -

**How to test**  
Check that the component still behaves as expected.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


nathalielindqvist (approved):
LGTM :)

PR #919, opened 2023-03-08
fix(datetime): update prop and default prop value to align with control

nathalielindqvist:
**Describe pull-request**  
Aligned prop and default prop value with control instead of leaving the default blank.

**Solving issue**  
Fixes: [DTS-1066](https://tegel.atlassian.net/jira/software/projects/DTS/boards/1?selectedIssue=DTS-1066)

**How to test**  
1. Go to Storybook link below
2. Check in Datetime -> All substories
3. Check that the component still changes in size when choosing different options of the "Size" control
4. Check the default value of the "size" prop in the Notes tab


mJarsater (approved):
👍

PR #932, opened 2023-03-13
chore(react-sandbox): responsive data table example

theJohnnyMe:
**Describe pull-request**  
Testing data-table component responsive feature in React sandbox

**Solving issue**  
There was a question from a user in the Development channel.

**Good to know**

- Separate page  was created to make tests for data table
- there are two different commits on how to deal with the responsiveness of data-table, please take a look on them
- this code relies on using the latest "@scania/components"

### DIFF
--- a/sandbox/react/src/App.js
+++ b/sandbox/react/src/App.js
@@ -4,6 +4,7 @@ import Header from './components/header/Header';
 import Form from './components/form/Form';
 import Home from './components/home/Home';
 import Footer from './components/footer/Footer';
+import DataTable from './components/dataTable/dataTable';
 
 function App() {
   const [expand, setExpand] = useState(false);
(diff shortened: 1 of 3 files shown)


mJarsater:
Could you commit the package.json also? So we get the latest @scania/components in the sandbox

theJohnnyMe:
> Could you commit the package.json also? So we get the latest @scania/components in the sandbox

Absolutely!

theJohnnyMe:
@mJarsater  - I did remove node_modules and run installation again. In this one, we use "latest" for all Scania packages. I can also specify 4.10.6 which is the last one released to be 100% sure, what do you think?

mJarsater:
> @mJarsater - I did remove node_modules and run installation again. In this one, we use "latest" for all Scania packages. I can also specify 4.10.6 which is the last one released to be 100% sure, what do you think?

I would say we go with latest, that way we don't have to think about maintaining it in the future 😄

theJohnnyMe:
> > @mJarsater - I did remove node_modules and run installation again. In this one, we use "latest" for all Scania packages. I can also specify 4.10.6 which is the last one released to be 100% sure, what do you think?
> 
> I would say we go with latest, that way we don't have to think about maintaining it in the future 😄

Then we are all good then, that one is in place already.

mJarsater (approved):
✅

PR #934, opened 2023-03-14
fix(card): introduce correct colors for icon

nathalielindqvist:
**Describe pull-request**  
Introduced color variables for card icon in dark and light mode. The color values will be updated once the updates to Figma UI-kit have been finalised. This is a first idea, please let me know if there's a better way to do this.

**Solving issue**  
Fixes: [DTS-949](https://tegel.atlassian.net/jira/software/projects/DTS/boards/1?selectedIssue=DTS-949)

**How to test**  
1. Go to Storybook link below
2. Check in Card -> Native and Web Component
3. Check that the arrow icon is grey-300 in light mode
4. Check that the arrow icon is grey-700 in dark mode

[DTS-949]: https://tegel.atlassian.net/browse/DTS-949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tegel/src/components/card/card.stories.tsx
+++ b/tegel/src/components/card/card.stories.tsx
@@ -121,7 +121,7 @@ export default {
     bodyImg: false,
     bodyText: '',
     bodyDivider: false,
-    cardBottom: '<sdds-icon style="font-size: 20px;" name="arrow_right"></sdds-icon>',
+    cardBottom: '<sdds-icon style="font-size: 20px; color: var(--sdds-card-icon-color);" name="arrow_right"></sdds-icon>',
     clickable: false,
   },
 };
--- a/tegel/src/components/card/card.stories.tsx
+++ b/tegel/src/components/card/card.stories.tsx
@@ -121,7 +121,7 @@ export default {
     bodyImg: false,
     bodyText: '',
     bodyDivider: false,
-    cardBottom: '<sdds-icon style="font-size: 20px; color: var(--sdds-card-icon-color);" name="arrow_right"></sdds-icon>',
+    cardBottom: '<sdds-icon style="font-size: 20px;" name="arrow_right"></sdds-icon>',
     clickable: false,
   },
 };
--- a/tegel/src/components/card/sdds-card.scss
+++ b/tegel/src/components/card/sdds-card.scss
@@ -77,7 +77,7 @@
   slot[name='card-bottom']::slotted(*) {
     display: flex;
     padding: 16px;
-
+    color: var(--sdds-card-icon-color);
   } 
 }
 
--- a/tegel/src/components/card/sdds-card.scss
+++ b/tegel/src/components/card/sdds-card.scss
@@ -78,10 +78,7 @@
     display: flex;
     padding: 16px;
 
-  }   
-  slot[name='card-bottom']::slotted(sdds-icon) {
-    color: var(--sdds-card-icon-color);
-  }   
+  } 
 }
 
 button {
--- a/tegel/src/components/card/sdds-card.scss
+++ b/tegel/src/components/card/sdds-card.scss
@@ -77,7 +77,11 @@
   slot[name='card-bottom']::slotted(*) {
     display: flex;
     padding: 16px;
-  }
+
+  }   
+  slot[name='card-bottom']::slotted(sdds-icon) {
+    color: var(--sdds-card-icon-color);
+  }   
 }
 
 button {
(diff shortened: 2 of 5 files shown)


tegel/src/components/card/sdds-card.scss
```diff
@@ -77,7 +77,11 @@
   slot[name='card-bottom']::slotted(*) {
     display: flex;
     padding: 16px;
-  }
+
+  }   
+  slot[name='card-bottom']::slotted(sdds-icon) {
+    color: var(--sdds-card-icon-color);
+  }
```
mJarsater, line 84: I don't think this one will work the way we want to. The ::slotted-pseudoselector isn't good enought to pick up the sdds-icon within it. 

Question: should icons always have the same color on the card no matter where it's placed? In that case we could just target the sdds-icon element and set the color directly on it.

mJarsater: i tried checkig out this branch and the changing it to red. Then I swapped the "Content of the bottom of the card" to this: `<div slot="card-bottom">
This i some text. 
<sdds-icon  style="font-size: 20px;" name="arrow_right"></sdds-icon>
</div>`

I would expect the text to be black and the icon to be red. But that does not work.

nathalielindqvist: Hmm yeah that's what I was afraid of. I agree that the ::slotted psudoselector is not stable enough for this, but I don't know what could be a better alternative. Simply targeting the sdds-icon doesn't work because of the slot, at least not when I try it out. Does it need to be a slot?

mJarsater: Yes I would say it needs to be a slot. Since we don't know what people will want to put in those spaces. I am thinking though, that maybe we should skip the all together. Not control the color, and simply leave it up to the user. And in storybook we could just give an example of how they should do it. 

So in storybook we would have this as the bottom slot content:
<sdds-icon color="var(--sdds-grey-300)/var(--sdds-grey-700) - based on the current theme"  style="font-size: 20px;" name="arrow_right"></sdds-icon>

nathalielindqvist: That is absolutely an idea, I was thinking the same for a while. It does leave a lot of freedom to the user where they could insert an icon using the wrong color. Are we ok with that?

tegel/src/components/card/card.stories.tsx
```diff
@@ -121,7 +121,7 @@ export default {
     bodyImg: false,
     bodyText: '',
     bodyDivider: false,
-    cardBottom: '<sdds-icon style="font-size: 20px;" name="arrow_right"></sdds-icon>',
+    cardBottom: '<sdds-icon style="font-size: 20px; color: var(--sdds-card-icon-color);" name="arrow_right"></sdds-icon>',
```
mJarsater, line 124: Yeah I like this approach more. Let's see what everyone else thinks 👍🏻

timrombergjakobsson (approved):
I like @mJarsater suggestion, so for me it is a go!

tegel/src/components/card/sdds-card.scss
EmelieLitwin: You seem to have found a solution that you are happy with, but an alternative might be to put the style on the parent element and let the icon inherit the color? Not familiar enough with the component to know if that works and if it would effective something else that could be in the bottom.

theJohnnyMe: +1 on what Emelie said. sdds-icon will inherit parents' color prop. I had the same/similar issue with textfield and prefix/suffix icons.

mJarsater: The problem is that any icons should have a different color than any text in the card "bottom". We don't know where the want to place this icon, before or after some text so we cant have a specific slot/prop for it. So we need to be flexible enough for people to place any icon/text where they want to, but we also need to style them differently. And targeting just an icon within the slot isn't possible.

alexandereneroth: As I discussed with @mJarsater setting the color in the parent might be good enough in this case, as the Figma design of the Card doesn't show any state with text in that part of the Card.
The only things apart from an icon that should be there are buttons or links, which have their own color set already, so they wouldn't be affected.

If the user puts text there, they could override it with css

nathalielindqvist: Updates have been made, great discussion! Please take a second look to make sure everyone is happy :)

alexandereneroth (approved):
Nice! works well

PR #944, opened 2023-03-16
fix(radio button): removed default slot 

mJarsater:
**Describe pull-request**  
Removed default slot and added named label slot.

**Solving issue**  
Fixes: -

**How to test**  
1. Go to storyook link below
2. Check in Components -> Radio button -> Web Component
3. Check that the label works as intended.A

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


theJohnnyMe (changes requested):
Clicking on the label does emit an event, but it does not make the option selected visually. Please see if we can make a "bridge" function that will set the option to check if the slot is clicked or something similar.

mJarsater:
> Clicking on the label does emit an event, but it does not make the option selected visually. Please see if we can make a "bridge" function that will set the option to check if the slot is clicked or something similar.

Its only in the canvas tab that is doesn't work right?
I think this might be a storybook issue. It works the same way before we introduced the slot.

theJohnnyMe (approved):
LGTM

theJohnnyMe:
> > Clicking on the label does emit an event, but it does not make the option selected visually. Please see if we can make a "bridge" function that will set the option to check if the slot is clicked or something similar.
> 
> Its only in the canvas tab that is doesn't work right? I think this might be a storybook issue. It works the same way before we introduced the slot.

True that, switching between Canvas and Docs made it work all od a sudden. All good them. Lets remind to check it in dark/light demo page once we release it.

PR #949, opened 2023-03-17
fix(stepper): added correct border styling on native label in dark mode

mJarsater:
**Describe pull-request**  
Added the correct border styling for inactive step in dark mode. 

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1269

**How to test**  
1. Go to storybook
2. Check in Components -> Stepper -> Native
3. Check that the styling is correct in both modes.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/tegel/src/components/stepper/stepper.scss
+++ b/tegel/src/components/stepper/stepper.scss
@@ -196,11 +196,17 @@
 
     &--inactive {
       .sdds-stepper__step-icon {
-        border-color: rgb(0 21 51 / 16%);
+        border-color: var(--sdds-stepper-label-border-color-upcoming);
+        opacity: var(--sdds-stepper-label-text-opacity-upcoming);
+
+        &::after, &::before {
+          opacity: unset;
+          opacity: 0.38;
+        }
       }
 
       .sdds-stepper__step-icon-value {
-        opacity: 0.38;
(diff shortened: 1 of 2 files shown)


tegel/src/components/stepper/stepper.scss
```diff
@@ -196,11 +196,17 @@
 
     &--inactive {
       .sdds-stepper__step-icon {
-        border-color: rgb(0 21 51 / 16%);
+        border-color: var(--sdds-stepper-label-border-color-upcoming);
+        opacity: var(--sdds-stepper-label-text-opacity-upcoming);
+
+        &::after, &::before {
+          opacity: unset;
```
nathalielindqvist, line 203: Does the opacity need to be unset before setting another opacity?

nathalielindqvist (approved):
Nice job! 🚀 Just a small question, except for that LGTM :)

tegel/src/components/stepper/stepper.scss
mJarsater: Yeah it does for now :/ Other wise it's gonna get a opacity based on it's parent opacity. Which we don't want

PR #954, opened 2023-03-20
fix(header): dark/light mode colors regression

theJohnnyMe:
**Describe pull-request**  
Checking color values for light and dark modes of header.

*Solving issue**  

Fixes: DTS-718

**How to test**  
1. Check the open state of the App Launcher icon, it should be white background and dark tots on it.
2. Check mobile version of Avatar/Profil dropdown - it needs to have the same background color in regular and hover state as the rest of the items mobile menu.
3. Border/Vertical divider between header items is updated to reflect the value in Figma (grey 700 instead of grey 800)

### DIFF
--- a/tegel/src/components/header/header.scss
+++ b/tegel/src/components/header/header.scss
@@ -381,7 +381,6 @@ html,
       width: 20px;
       height: auto;
       transition: fill 150ms ease;
-      color: var(--sdds-header-app-launcher-btn-color);
     }
   }
 
@@ -869,10 +868,8 @@ html,
       width: 100%;
 
       &::before {
-        //@include type-style('headline-07');
         font: var(--sdds-headline-07);
         letter-spacing: var(--sdds-headline-07-ls);
-        color: var(--sdds-grey-958);
         content: 'Search';
         position: absolute;
         top: 50%;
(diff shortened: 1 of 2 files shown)


nathalielindqvist:
Not sure if this belongs in this PR, but I found a strange bug with the App Launcher dropdown menu. A small strip of the wrong color appears at the bottom of the menu. It only appears in a very specific state, just as the scroll function kicks in. It happens in both light and dark mode.

image

image

tegel/src/components/header/header.scss
```diff
@@ -1005,7 +1002,19 @@ html,
...
+      }
+
+      .sdds-nav__avatar-item {
+        border-bottom: 1px solid var(--sdds-header-mobile-nav-center-item-border-color);
+
+        .sdds-nav__avatar-item-core {
+          background-color: var(--sdds-header-mobile-nav-center-background);
+
+          &:hover {
+            background-color: var(--sdds-header-mobile-avatar--item-btn-background-hover);
```
nathalielindqvist, line 1015: Not sure if this is intentional, but the background color on hover is different on the user menu between the mobile and desktop menu. It only seems to be happening in dark mode. Seeing as the background color are different as well, that might be why?

image
image
image
image

alexandereneroth (approved):
Looks good to me 👍

tegel/src/components/header/header.scss
alexandereneroth: Issue created: DTS-1391

alexandereneroth:
> Not sure if this belongs in this PR, but I found a strange bug with the App Launcher dropdown menu. A small strip of the wrong color appears at the bottom of the menu. It only appears in a very specific state, just as the scroll function kicks in. It happens in both light and dark mode.
> 
> image
> 
> image

I think this is a separate issue, not related to light/dark mode.
It seems to only happen when you use the storybook light/dark mode switcher, after a page refresh the issues is gone.
Ill create a JIRA issue for this.

theJohnnyMe:
@nathalielindqvist and @alexandereneroth - Thanks for the review and comments. The bug in the app launcher seems to be in Storybook only. Sometimes when the height is changed, it works properly. Thanks for making ticket for it Alex and Nathalie for spotting it ;)

PR #956, opened 2023-03-20
feat(tabs): tab item

mJarsater:
**Describe pull-request**  
This PR introduces 3 new components. folder-, inline- and navigation-tabs-item. These are three generic components that can be used to add a react router Link or NavLink component for the tabs.

**Solving issue**  
Fixes: -

**How to test**  
2. Check in Componets -> Tabs -> Folder/Inline/Navigation tabs
3. Change the "Tab type" to item and try the tabs out.
4. Check that the styling i still correct.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


tegel/src/components/tabs/navigation-tabs/navigation-tabs-item/sdds-navigation-tabs-item.scss
```diff
@@ -0,0 +1,92 @@
...
+      cursor: pointer;
+
+      &::after {
+        width: 100%;
+      }
+    }
+  }
+
+  ::slotted(*:focus-visible) {
+    outline: 2px solid var(--sdds-blue-400);
```
timrombergjakobsson, line 41: These 2 lines almost feel like they could be a mixin? If we like mixins :). As they are repeated often.

timrombergjakobsson:
```suggestion
    @mixin custom-outline($color, $offset) {
       outline: 2px solid $color;
       outline-offset: $offset;
     }
```

mJarsater: Yes, already have a mixin for it actuallt @sdds-focus-state. I was playing around with it a bit and that's why it's not using it. Nice catch 🎣

nathalielindqvist:
This is probably not related directly to this PR, but when going down to mobile size on the Inline Tabs the button to skip-right works well but the button to skip-left only skips in very small increments, so you have to click a lot of times to get back to the start of the tabs. This does not happen with the other tabs versions. Seems like that's not supposed to happen? :)

mJarsater:
> This is probably not related directly to this PR, but when going down to mobile size on the Inline Tabs the button to skip-right works well but the button to skip-left only skips in very small increments, so you have to click a lot of times to get back to the start of the tabs. This does not happen with the other tabs versions. Seems like that's not supposed to happen? :)

Nice catch, should be fixed now. 🔨

nathalielindqvist (approved):
Really nice job! I like the more neutral alternative, feels more flexible 🥳 LGTM!

alexandereneroth (changes requested):
Lets keep this to work more on /test how the tab API could be

tegel/src/components/tabs/folder-tabs/folder-tabs.tsx
```diff
@@ -173,60 +202,15 @@ export class InlineTabs {
...
-      selectedTab: this.selectedTab,
-      selectedTabIndex: this.selectedTabIndex,
-    });
-  }
-
   render() {
     return (
       <Host class={`${this.modeVariant ? `sdds-mode-variant-${this.modeVariant}` : ''}`}>
         <div
+          role="list"
```
alexandereneroth, line 209: I think the role needs to be on the Host element, since the div is inside the shadow dom

alexandereneroth (changes requested):

PR #962, opened 2023-03-20
fix: set default mode variant color and bgcolor

alexandereneroth:
**Describe pull-request**  
It makes sense that if we change the text to darkmode we would also change the default background so the text is not invisible.
This PR addresses that

**How to test** 
Clone the branch, add a .sdds-mode-dark somewhere, and see the background change!
Optionally test it more or less thoroughly.


mJarsater (approved):
🦄

PR #972, opened 2023-03-22
fix(textfield): position of error helper

mJarsater:
**Describe pull-request**  
The textfield component was using two different slots for the error based on the state of the component. This lead to people updating the state in runtime using the wrong slot (without them knowing it.), since the slot doesn't react to prop changes. This should solve this issue.

**Solving issue**  
Fixes: -

**How to test**  
1. Check out the branch and run storybook for components.
2. Check in Components -> Textfield
3. Try out the helper text with different states and max length and so on.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


EmelieLitwin (approved):
Bug smasher #1! Good find!

timrombergjakobsson (approved):

PR #978, opened 2023-03-22
fix(sider-menu): correct color on header in side menu dark mode

timrombergjakobsson:
**Describe pull-request**  
Correct color for close icons header part. 

**Solving issue**  
Fixes: [#](https://tegel.atlassian.net/browse/DTS-1394)

**How to test**  
1.Go to Storybook link below
2. Check in Header -> Inline Menu, switch to dark mode
Resize the window so the menu items are moved into the side menu
Open the menu and see that the color is the same as Figma.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

**Screenshots** 
Should look like this below.
Skärmavbild 2023-03-22 kl  16 16 38

### DIFF
--- a/tegel/src/components/header/header-vars.scss
+++ b/tegel/src/components/header/header-vars.scss
@@ -103,11 +103,11 @@
   --sdds-header-background: var(--sdds-blue-900);
   --sdds-header-mob-menu-open-color: var(--sdds-white);
   --sdds-header-mob-menu-open-background: var(--sdds-grey-958);
-  --sdds-header-mob-menu-open-border: var(--sdds-blue-700);
+  --sdds-header-mob-menu-open-border: var(--sdds-grey-868);
   --sdds-nav-item-background: var(--sdds-grey-800);
   --sdds-nav-item-background-hover: var(--sdds-grey-700);
   --sdds-nav-item-background-active: var(--sdds-grey-700);
-  --sdds-nav-item-border-color: var(--sdds-blue-700);
+  --sdds-nav-item-border-color: var(--sdds-grey-868);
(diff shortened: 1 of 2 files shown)


timrombergjakobsson:
> Screenshot 2023-03-22 at 16 31 16
> 
> While not technically requested by the JIRA issue, it would be nice if we could fix the borders I hightlighted in red here too ☝️

Sure, I can fix them as well.

tegel/src/components/header/header-vars.scss
```diff
@@ -102,12 +102,12 @@
...
   --sdds-header-mob-menu-open-color: var(--sdds-white);
-  --sdds-header-mob-menu-open-background: var(--sdds-blue-800);
-  --sdds-header-mob-menu-open-border: var(--sdds-blue-700);
+  --sdds-header-mob-menu-open-background: var(--sdds-grey-958);
+  --sdds-header-mob-menu-open-border: var(--sdds-grey-868);
   --sdds-nav-item-background: var(--sdds-grey-800);
   --sdds-nav-item-background-hover: var(--sdds-grey-700);
   --sdds-nav-item-background-active: var(--sdds-grey-700);
-  --sdds-nav-item-border-color: var(--sdds-blue-700);
+  --sdds-nav-item-border-color: var(--sdds-grey-868);
```
alexandereneroth, line 110: Unfortunately this change changed more than intended:

Before:
Screenshot 2023-03-22 at 17 13 17

After:
Screenshot 2023-03-22 at 17 13 08

alexandereneroth (changes requested):
see comment

alexandereneroth (approved):
Nice work!

Issue #975, opened 2023-03-22
[Bug report]: sddsModal throws error when using openModal and not passing a selector


toliv3:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/sdds-website/issues) reported.
- [X] I have read the contribution [guidelines](https://github.com/mmexvr/sdds/blob/master/CONTRIBUTING.md).
- [X] I have checked the lastest version if the bug exist there. See all avaliable packages at [npmJS.com](https://www.npmjs.com/search?q=%40scania)
- [X] I have made sure that the issue comes from SDDS packages by removing SDDS from my project as a test.
- [X] I have followed the installation guide.

### Package

@scania/components

### Package versions

@scania/components@4.10.7

### Browser

Chrome

### Framework

Angular

### Version

Angular 15

### Reproduction steps

When using sdds-modal with the openModal/closeModal methods, there is no need to pass a selector for the modal to wire events to. If we don't pass a random selector the component will throw errors on the console. (just add a null check when initializing the component for the selector prop)

### Code example

<div>....

### Screenshots

![image](https://user-images.githubusercontent.com/111114201/226903079-53c70ef5-d8a4-4826-8cec-32cdbbc7250e.png)

### Expected behaviour

Component should not throw errors when selector is null/empty.

### Console errors

Add warning/error message

### Contact information

thiago.oliveira@scania.com

PR #980, opened 2023-03-23
fix(modal): added null check to avoid empty selector

mJarsater:
**Describe pull-request**  
If the user used the modal methods to contol the open/closed state of the modal it would render an error in the console where we did a querySelectAll() - with an empty argument. This PR adds a simple null check to avoid this.

**Solving issue**  
Fixes: #975

**How to test**  
1. Check out branch.
2. Check in Components -> Modal
3. Remove the selector from one of the modal and open it with it's class method.
4. Check that there are no errors in the console.

### DIFF
--- a/components/src/components/modal/modal.tsx
+++ b/components/src/components/modal/modal.tsx
@@ -44,22 +44,22 @@ export class Modal {
 
   componentDidLoad() {
     let targets;
-    if (!this.selector) {
+    if (this.selector) {
       targets = document.querySelectorAll(this.selector);
+
+      // If the modal doesn't have a selector to be triggered
+      if (!targets) {
+        console.warn('No prop for modal targeted, please add selector attribute');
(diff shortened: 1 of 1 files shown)


components/src/components/modal/modal.tsx
```diff
@@ -43,7 +43,10 @@ export class Modal {
   }
 
   componentDidLoad() {
-    const targets = document.querySelectorAll(this.selector);
+    let targets;
+    if (!this.selector) {
```
timrombergjakobsson, line 47: Hmm, should it not be `if (this.selector)`? Should we not check if it is truthy instead?

mJarsater, line 47:
```suggestion
    if (this.selector) {
```

timrombergjakobsson (approved):

PR #984, opened 2023-03-27
feat(accordion): cancelable events

mJarsater:
**Describe pull-request**  
This PR enables for users to cancel the sddsToggle event that is emitted whenever a accordion-item is toggled. It also adds a public method for the users to trigger the event with. Lastly I added a "accordion-item-header" slot that can be used instead of the 'header' prop currently on the component.

**Solving issue**  
Fixes: [DTS-1083](https://tegel.atlassian.net/browse/DTS-1083)

**How to test**  
2. Check in Components -> Accordion
3. Check the readme for the event and the public method.
4. Check out the branch
5. Try preventing an event
6. See it fail to execute

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

[DTS-1083]: https://tegel.atlassian.net/browse/DTS-1083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tegel/src/components/accordion/accordion-item/accordion-item.tsx
+++ b/tegel/src/components/accordion/accordion-item/accordion-item.tsx
@@ -24,8 +24,9 @@ export class AccordionItem {
   /** Method for toggeling the expanded state of the accordion item. */
   @Method()
   async toggleAccordionItem() {
+    // This is negated in order to emit the value the accordion item will have after it has expanded/redacted.
     const event = this.sddsToggle.emit({
-      expanded: this.expanded,
+      expanded: !this.expanded,
     });
     if (!event.defaultPrevented) {
       this.expanded = !this.expanded;
--- a/tegel/src/components/accordion/accordion-item/accordion-item.tsx
+++ b/tegel/src/components/accordion/accordion-item/accordion-item.tsx
@@ -21,9 +21,9 @@ export class AccordionItem {
   /** When true 16px on right padding instead of 64px */
   @Prop() paddingReset: boolean = false;
 
-  /**  */
+  /** Method for toggeling the expanded state of the accordion item. */
   @Method()
-  toggleAccordionItem() {
+  async toggleAccordionItem() {
     const event = this.sddsToggle.emit({
       expanded: this.expanded,
     });
--- a/tegel/src/components/accordion/accordion-item/accordion-item.tsx
+++ b/tegel/src/components/accordion/accordion-item/accordion-item.tsx
@@ -1,4 +1,4 @@
-import { Component, Event, EventEmitter, h, Host, Prop } from '@stencil/core';
+import { Component, Event, EventEmitter, h, Host, Method, Prop } from '@stencil/core';
 
 @Component({
   tag: 'sdds-accordion-item',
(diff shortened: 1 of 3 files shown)


tegel/src/components/accordion/accordion-item/accordion-item.tsx
```diff
@@ -21,19 +21,27 @@ export class AccordionItem {
...
+  /** Method for toggeling the expanded state of the accordion item. */
+  @Method()
+  async toggleAccordionItem() {
+    const event = this.sddsToggle.emit({
+      expanded: this.expanded,
+    });
+    if (!event.defaultPrevented) {
+      this.expanded = !this.expanded;
+    }
+  }
```
alexandereneroth, line 33:
```suggestion
  async toggleAccordionItem() {
  	const toggledExpanded = !this.expanded;
    const event = this.sddsToggle.emit({
      expanded: toggledExpanded,
    });
    if (!event.defaultPrevented) {
      this.expanded = toggledExpanded;
    }
  }
```
This code emits with the expanded state before it is toggled, so it will show the old value, not the new one, which might cause confusion.

In the suggestion i made it so the event emits with the toggled value, but it is not set if default is prevented.

tegel/src/components/accordion/accordion-item/accordion-item.tsx
```diff
@@ -21,19 +21,27 @@ export class AccordionItem {
   /** When true 16px on right padding instead of 64px */
   @Prop() paddingReset: boolean = false;
 
-  /** Fires after the accordion item is closed or opened, emitting the value (as boolean) of the current state of the accordion */
+  /** Method for toggeling the expanded state of the accordion item. */
```
alexandereneroth, line 24:
```suggestion
  /** Method for toggling the expanded state of the accordion item. */
```

alexandereneroth (changes requested):

tegel/src/components/accordion/accordion-item/accordion-item.tsx
mJarsater: Yeah I'm not sure how to handle these really. Ideally I would not want to emit a value that's "false".

The code above could just be: 
` async toggleAccordionItem() {
    const event = this.sddsToggle.emit({
      expanded: !this.expanded,
    });
    if (!event.defaultPrevented) {
      this.expanded = !this.expanded;
    }
  }`
  
  We don't really need the temp value. Would should leave a comment explaining it for future reference though

alexandereneroth: Yea ChatGPT gave me that as an example to fix it, but then I asked to compare it to the one with the variable it gave me this 🤷 

> Both code snippets achieve the same goal of preventing the expanded state from being toggled if the event default is prevented. The code you provided is a bit more explicit, and it has some advantages:

>1. Readability: By creating a separate variable (nextExpanded) to store the opposite value of the current expanded state, the code becomes more readable and easier to understand.

>2. Maintainability: With a separate variable, it's easier to modify the code in the future if there's any change in the logic of how the next state is calculated.

>3. Debugging: Having a separate variable makes it easier to debug the code if any issues arise since you can directly inspect the nextExpanded value.

But don't think it really matters which way TBH

alexandereneroth (approved):
👍

PR #990, opened 2023-03-29
feat(utils): appendHiddenInput

mJarsater:
**Describe pull-request**  
Adds a utils method for appending a hidden input for form elements. This can be used by our own input element that needs has a shadowDOM.

**Solving issue** 
Fixes: #


EmelieLitwin:
Really nice! 🙌  Maybe the parameter "attributes", could be called something like "otherAttributes" or something like that just to make it extra clear and easy to follow.

EmelieLitwin (approved):
🥳 Nice

PR #997, opened 2023-03-30
fix(data-table): made all events non-cancelable

mJarsater:
**Describe pull-request**  
Made all events non-cancelable in the data-table. If the user want to implement their own logic for certain events there are props in place for that. See: enable-client-pagination.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1092


nathalielindqvist (approved):
👌

PR #998, opened 2023-04-03
fix(global): bg and color - default to light mode

mJarsater:
**Describe pull-request**  
This PR enables light mode by default.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1477

**How to test**  
2. Check in Components -> Data-table -> WC
3. Check that primary vs secondary is correct.
4. Check the "root" of the HTML.
5. See that it has the correct bg and color based on theme.

### DIFF
--- a/tegel/src/global/global.scss
+++ b/tegel/src/global/global.scss
@@ -86,17 +86,16 @@ body {
   flex: 1;
 }
 
-html,
 .sdds-mode-light {
   color: var(--sdds-grey-958);
   background-color: var(--sdds-white);
 
-  & .sdds-mode-variant-primary {
+  &.sdds-mode-variant-primary {
     color: var(--sdds-grey-958);
     background-color: var(--sdds-white);
   }
 
-  & .sdds-mode-variant-secondary {
+  &.sdds-mode-variant-secondary {
     color: var(--sdds-grey-958);
     background-color: var(--sdds-grey-50);
   }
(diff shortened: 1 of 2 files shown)


nathalielindqvist (approved):
LGTM :)

EmelieLitwin:
Not really part of the PR but in most components primary/secondary mode variant variables have been handled like this:
image

E.i. all the background colors etc have been declared at the same place above. Should we do it like that in the data table as well?

mJarsater:
> Not really part of the PR but in most components primary/secondary mode variant variables have been handled like this: image
> 
> E.i. all the background colors etc have been declared at the same place above. Should we do it like that in the data table as well?

Yes, I think that would be a good approach 👍

tegel/src/global/global.scss
```diff
@@ -86,17 +86,16 @@ body {
   flex: 1;
 }
 
-html,
 .sdds-mode-light {
```
alexandereneroth, line 89:
```suggestion
:root,
.sdds-mode-light {
```

alexandereneroth (changes requested):
I think the issue is rather that the background style is set on the :host of the webcomponent (&lt;sdds-table-body-row>), I think we should not style :host as styles put on :host are outside the shadow dom and are not protected (as this issue demonstrates).
It would be better if we put the background on an inner element inside :host instead.

PR #1008, opened 2023-04-11
fix(footer): renamed files and created a footer-item instead of link

mJarsater:
**Describe pull-request**  
Renamed and cleaned up files. Also created a footer-item instead of footer-link. To allow users to slot in any link.

Note: Breakpoint will remain unchanged. The footer breaks as 992px.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-722

**How to test**  
2. Check in Components -> Footer ->Web Component
3. Check that the footer breaks as 992px.


nathalielindqvist (approved):
LGTM!

PR #1012, opened 2023-04-14
feat(datetime): removed shadowDOM

mJarsater:
**Describe pull-request**  
Removed shadowDOM on the datetime to allow it to work within a form. Had to unset the background on the wrapping element with the class `sdds-form-datetime` to not have its background being set by our global styling.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1462

**How to test**  
1. Go to Components -> Datetime
2. Check that the styling is still correct for all the variants.


nathalielindqvist (approved):
LGTM :)

Issue #257, opened 2021-12-20
Bug - Footer above the sidebar overlay in mobile


mmexvr:
**Describe the bug**  
_A clear and concise description of what the bug is._
Footer component is over the navbar overlay for mobile view. 

**To Reproduce**   
_Steps to reproduce the behavior and be very detailed_  
1. Run angular with side-menu
2. Change resolution, to mobile
3. Scroll down to footer
4. open the sidebar menu
5. see that the footer is above the overlay 

**Expected behavior**  
_A clear and concise description of what you expected to happen._
Footer not having more z-index then side-menu overlay

**Screenshots**  
_If applicable, add screenshots to help explain your problem._
![Screenshot 2021-12-20 at 11 27 09](https://user-images.githubusercontent.com/35451568/146753406-35936c90-f53d-4181-841f-1efa37527626.png)

**Version of SDDS or which package**  
 _E.g 1.2.3 or @scania/theme-light v2.3.4_

**Application link**  
_Provide link if possible_  
- [https://example.com](/)

**System information:**  
_Please complete the following information, don't forget version_
 - Framework: angular 11
 - Device: Mac
 - OS: MacOS
 - Browser: chrome

**Additional context**   
_Add any other context about the problem here._

theJohnnyMe:
@alexlioliopoulos and @victorscania  - I am tagging you just because we will have a new Side Menu out soo and will integrate it together with the new Footer on the website, so might face this issue. Therefore, let's think of a quick fix for this until we have a z-index workshop and we sort it out in a smarter way.

PR #1033, opened 2023-04-26
fix(components/header): fix footer overlapping old header sidemenu

alexandereneroth:
**Describe pull-request**  
- In @scania/components: Fixes old footer overlapping side menu.
- In angular sandbox: Changes header so that the bug can be reproduced.

**Solving issue**  
Fixes: #257, https://tegel.atlassian.net/browse/DTS-787 

**How to test**  
1. Checkout the branch locally
1. Make sure you are on node version 16 and you ran npm install in the appropriate folders
1. Go to the sdds root folder
1. Run this and it should hopefully just work™️
```
cd ./components && npm run build && npm link && cd ../sandbox/angular && npm link @scania/components && npm start
```
5. Once the dev server is running in the angular sandbox, go to the path `/short` in the URL bar
6. Open the sidemenu and see it overlapping the footer! 🏁 

**Screenshots**  
Previous issue:
Screenshot 2023-04-26 at 17 03 42

After fix:
Screenshot 2023-04-26 at 17 04 37

**Additional context**  
Could not reproduce the original issue exactly but it is pretty old. This fix should fix it nonetheless.


components/src/components/header/header.scss
```diff
@@ -39,11 +37,12 @@ html,
   height: 64px;
   width: 100%;
   z-index: 3;
+  position: relative;
```
mJarsater, line 40: Is this the only "real" change?

mJarsater: If so, im not sure I understand how that fixes our issue. Would love a quick walkthrough

alexandereneroth: Its basically this: z-index doesnt do anything on position:static (which is the default) 😃 
If you read here, you'll see its supposed to be set on a "positioned" element: https://developer.mozilla.org/en-US/docs/Web/CSS/z-index

mJarsater: Aaah! Lovely!

mJarsater (approved):
🚀

PR #1036, opened 2023-05-09
fix(slider): keyboard nav snapping

alexandereneroth:
**Describe pull-request**  
Fix slider keyboard adjustment not working properly when "snap to ticks" is enabled.

**How to test**
- Go to the slider story
- Turn on ticks and snap to ticks
- Operate the slider with your arrow keys and see that it all works
- Operate the slider with your mouse and see that it all works
- Try to break it
- Celebrate

**Solving issue**  
#DTS-1570

**Additional context**  
I think this component badly needs a refactor/rewrite if its something we want to spend time on in the future. Can barely follow what I did myself.


alexandereneroth:
> LGTM! Nice fix man! ;) The only thing that might be nice to have is to add an arrow Up/Down as well as +/- keyboard navigation as we have it in Tegel repo.

Thanks!
Added the extra keyboard navigation keys =)

PR #1043, opened 2023-05-22
docs(readme): updated readme for theme to only include current state

mJarsater:
Updated the readme for sdds/theme to only include current state.

Fixes: https://tegel.atlassian.net/browse/DTS-1718


alexandereneroth (approved):
👍

